PR #105, opened 2018-05-23
Add pagination to audit trails

vicmgs:


assets/js/components/dashboard/Audit.jsx
```diff
@@ -62,21 +108,31 @@ class Audit extends Component {
...
+      pageSize,
+      pageNumber
     }
   }
 `
 
 const queryOptions = {
   options: props => ({
-    fetchPolicy: 'network-only'
+    fetchPolicy: 'network-only',
```
allenan: what's this?

vicmgs: makes it so that queries don't get cached. Without this, new actions don't populate unless you refresh the browser. Making it network-only makes it more real-time

allenan: ok that's fine for now

PR #107, opened 2018-05-23
Upgrade react and material UI

vicmgs:


allenan:
nice! do we need to make any changes to the lifecycle methods for the react upgrade? or are we already using the correct methods?

vicmgs:
good point, let me look

PR #111, opened 2018-05-29
Add paginated event table to gateway and channel

vicmgs:


vicmgs:
Do we want to pull gateway data through gql? since we do not have groups for gateways

allenan:
I think we should move toward doing all frontend querying through gql. ideally we'd be able to replace redux with it

vicmgs:
updated gateway and channel show pages to use gql to pull data. 
Next steps after this PR:

1. add gql subscription to channel + gateway
2. paginate resource index pages
3. get event graphs to work with gql

allenan:
looks good! 👍 On an unrelated note, I think we might want to move away from the relay-style connections (we're just using them for groups right now, and I'm not sure how much utility we get from them given we can't use it for our style of pagination) But we can save that for another PR

PR #112, opened 2018-05-30
Add event subscriptions for channel + gateway, link packet graphs to gql

vicmgs:


lib/console_web/controllers/event_controller.ex
```diff
@@ -68,5 +68,11 @@ defmodule ConsoleWeb.EventController do
     if event.device do
       Absinthe.Subscription.publish(ConsoleWeb.Endpoint, event, event_added: event.device.id)
     end
+    if event.gateway do
+      Absinthe.Subscription.publish(ConsoleWeb.Endpoint, event, event_added: event.gateway.id)
+    end
+    if event.channel do
+      Absinthe.Subscription.publish(ConsoleWeb.Endpoint, event, event_added: event.channel.id)
```
allenan: do we need some more context here? we're just using id it looks like

assets/js/components/events/EventsTablePaginated.jsx
```diff
@@ -174,8 +174,8 @@ const query = gql`
 
 // TODO create event fragment?
```
allenan: we're now using the event fields in a few places. could be worth looking into creating a fragment for it: https://www.apollographql.com/docs/react/advanced/fragments.html

assets/js/components/common/PacketGraph.jsx
```diff
@@ -83,17 +87,13 @@ class PacketGraph extends Component {
...
-      (this.props.data[0] && this.state.data.datasets.length == 0) ||
-      (this.props.data[0] && prevProps.data[0] && this.props.data[this.props.data.length-1].id !== prevProps.data[prevProps.data.length-1].id)
-    ) {
+    if (!prevProps.data.recentEvents && recentEvents) {
       clearInterval(this.chartUpdateInterval)
-      this.updateChart(this.props.data)
+      this.updateChart(recentEvents)
       this.chartUpdateInterval = setInterval(() => {
-        this.updateChart(this.props.data)
+        this.updateChart(recentEvents)
```
allenan: is this just polling that endpoint? could we use subscriptions here so we hit the gql endpoint once, grab all recent events, and then when new ones come in, add to the graph?

vicmgs: updateChart just transforms props to state for the chart Component, it doesn't hit the endpoint.

lib/console_web/controllers/event_controller.ex
vicmgs: don't think so? the if statement should router the new event to the proper context. Even if the event belongs to more than 1 context, we should publish for each context as well right?

vicmgs:
Fragmented out events gql stuff.

Also, added code for subscription in PacketGraph.jsx

assets/js/graphql/events.js
```diff
@@ -0,0 +1,21 @@
+import gql from 'graphql-tag';
+
+export const EventFragment = gql`
```
allenan: I usually think of camel-cased variables in javascript to mean classes. should this be a screaming snake case like `EVENTS_SUBSCRIPTION`?

lib/console_web/controllers/event_controller.ex
allenan: the only difference i'm seeing between these 3 publishes is the `event_added: event.channel.id` vs `event_added: event.gateway.id` which is the id of the event parent.

lib/console_web/schema/schema.ex
```diff
@@ -124,14 +111,21 @@ defmodule ConsoleWeb.Schema do
...
   end
 
   subscription do
     field :event_added, :event do
-      arg :device_id, :string
+      arg :context_id, :string
 
       config fn args, _ ->
-        {:ok, topic: args.device_id}
+        {:ok, topic: args.context_id}
```
allenan: yeah the subscription is just taking a context_id, so maybe we should make the topic like `"#{args.context_name}/#{args.context_id}"`

PR #113, opened 2018-05-31
Remove relay

vicmgs:


lib/console_web/schema/schema.ex
```diff
@@ -75,6 +34,33 @@ defmodule ConsoleWeb.Schema do
     field :latitude, :decimal
   end
 
+  object :device do
+    field :id, :id
+    field :_id, :string, resolve: &internal_id/2
```
allenan: so i think relay was what was creating a different kind of `id`, which is why I needed to create `_id` as a way to access the actual object id. with it removed, `id` should now be the actual object id so we might not need `_id` anymore

vicmgs: ah ok

vicmgs: should be good

PR #117, opened 2018-06-04
Add gql subscription to devices/channels/gateways

vicmgs:
+ Refactor table footer pagination
+ Create fragments for device/channel/gateway

should device/channel/gateway index graphql queries not be cached? Currently, creating a new channel using the button and clicking back to the channelIndex page causes the new channel to not show up until we refresh the page

### DIFF
--- a/lib/console_web/schema/schema.ex
+++ b/lib/console_web/schema/schema.ex
@@ -135,5 +135,29 @@ defmodule ConsoleWeb.Schema do
         {:ok, topic: "#{args.context_name}/#{args.context_id}"}
       end
     end
+
+    field :device_added, :device do
+      arg :team_id, :string
+
+      config fn args, _ ->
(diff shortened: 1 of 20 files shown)


allenan:
looks good. I have an idea for a further refactor into something like `<PaginatedTable />` since we repeat a lot of the paging logic. But I can do that in my cleanup branch once this is merged

lib/console_web/schema/schema.ex
```diff
@@ -135,5 +135,29 @@ defmodule ConsoleWeb.Schema do
         {:ok, topic: "#{args.context_name}/#{args.context_id}"}
       end
     end
+
+    field :device_added, :device do
+      arg :team_id, :string
```
allenan: hm yeah i think we need to find a way to validate that the current token has access to subscribe to a given channel based on the token's current team claim.

PR #118, opened 2018-06-05
UI Cleanup

allenan:
- restyle members table to have a title and action button in top right (this makes room for pagination below)
- move the map on gateway show page into the right half of the details card
- change the audit table default page size to 5
- make gateway maps its own route and make it searchable
- convert to using the decorator pattern for `withStyles`, `withTheme`, `connect` and `graphql`


assets/js/components/Home.jsx
```diff
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
 import { connect } from 'react-redux';
 
+@connect(mapStateToProps, {})
```
vicmgs: nicee

PR #120, opened 2018-06-06
Add geocoding to gateway create

vicmgs:
Yet to do:
1. Update gateway location by re-geocoding if geocoding during gateway creation fails, IE location field == api_call_failed

### DIFF
--- a/assets/js/components/common/BlankSlate.jsx
+++ b/assets/js/components/common/BlankSlate.jsx
@@ -29,10 +29,6 @@ const styles = theme => ({
 
 @withStyles(styles)
 class BlankSlate extends Component {
-  constructor(props) {
-    super(props)
-  }
-  
   render() {
     const { classes, title, subheading } = this.props
 
--- a/assets/js/components/common/BlankSlate.jsx
+++ b/assets/js/components/common/BlankSlate.jsx
@@ -29,18 +29,22 @@ const styles = theme => ({
 
 @withStyles(styles)
 class BlankSlate extends Component {
+  constructor(props) {
+    super(props)
+  }
+  
   render() {
-    const { classes } = props
+    const { classes, title, subheading } = this.props
 
     return (
       <div className={classes.main}>
         <Icon className={classes.icon} />
         <Typography variant="display1" className={classes.title}>
-          {props.title}
+          {title}
         </Typography>
 
         <Typography component="p" className={classes.subheading}>
-          {props.subheading}
+          {subheading}
         </Typography>
       </div>
     )
--- a/lib/console/gateways/gateway.ex
+++ b/lib/console/gateways/gateway.ex
@@ -12,6 +12,7 @@ defmodule Console.Gateways.Gateway do
     field :longitude, :decimal
     field :mac, :string
     field :name, :string
+    field :location, :string
     field :public_key, :binary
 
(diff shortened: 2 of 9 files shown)


lib/console_web/controllers/gateway_controller.ex
```diff
@@ -66,4 +68,25 @@ defmodule ConsoleWeb.GatewayController do
 
     Absinthe.Subscription.publish(ConsoleWeb.Endpoint, gateway, gateway_added: "#{gateway.team.id}/gateway_added")
   end
+
+  defp geocodeLatLng(%{"latitude" => lat, "longitude" => lng}) do
```
allenan: this might be worth at least moving to the Gateways context, if not to its own module

lib/console_web/controllers/gateway_controller.ex
```diff
@@ -18,7 +18,9 @@ defmodule ConsoleWeb.GatewayController do
   def create(conn, %{"gateway" => gateway_params}) do
     current_user = conn.assigns.current_user
     current_team = conn.assigns.current_team
-    gateway_params = Map.merge(gateway_params, %{"team_id" => current_team.id})
+    location = geocodeLatLng(gateway_params)
+
+    gateway_params = Map.merge(gateway_params, %{"team_id" => current_team.id, "location" => location})
```
allenan: this should be in a changeset so that when lat/lng are changed, the location is changed as well

assets/js/components/common/BlankSlate.jsx
```diff
@@ -29,18 +29,22 @@ const styles = theme => ({
 
 @withStyles(styles)
 class BlankSlate extends Component {
+  constructor(props) {
```
allenan: is this constructor needed for anything?

vicmgs: oh right i dont need the constructor. It was broken before because we didnt' have this.props, just props

lib/console_web/controllers/gateway_controller.ex
vicmgs: good point

lib/console/gateways/gateway.ex
```diff
@@ -24,8 +25,17 @@ defmodule Console.Gateways.Gateway do
...
-    |> cast(attrs, [:name, :mac, :public_key, :latitude, :longitude, :team_id, :location])
+    |> cast(attrs, [:name, :mac, :public_key, :latitude, :longitude, :team_id])
     |> validate_required([:name, :mac, :public_key, :latitude, :longitude, :team_id])
+    |> putGeocodingLocation()
     |> unique_constraint(:mac)
   end
+
+  defp putGeocodingLocation(changeset) do
+    case changeset do
+      %Ecto.Changeset{valid?: true, changes: changes} ->
```
allenan: you can pattern match latitude and longitude out of `changes:`

allenan: the reason to do that is so that the case won't match if lat/lng aren't present

vicmgs:
cool, thanks!

lib/console/gateways/gateway.ex
vicmgs: they should get filtered out by the validate_required from line 29 right?

PR #125, opened 2018-06-11
Refactor teamShow page to use paginated tables + clean up userCan

vicmgs:


assets/js/components/teams/TeamShow.jsx
```diff
@@ -130,38 +114,54 @@ class TeamShow extends Component {
...
+      label: "Access View",
+      content: accessView,
+      path: "/teams/access",
+    }, {
+      label: "Activity View",
+      content: activityView,
+      path: "/teams/activity",
+    }]
+
+    if (userCan("view", "auditTrails", null, this.props.user)) {
```
vicmgs: For the tabs, I need userCan to be hooked up to the store and get updates, or else the component gets the original null value for the user role. I was trying to use the UserCan component to achieve this behavior but couldn't think of a clean way since I just need to return true or false from userCan. 

Let me know what you think

assets/js/util/abilities.js
```diff
@@ -1,7 +1,7 @@
 import { store } from '../store/configureStore';
 
-const userCan = (action, itemType, item) => {
-  const user = store.getState().user
+const userCan = (action, itemType, item, user) => {
```
vicmgs: Same issue as above, this is temporary. We could just not import the store into this file, and use mapStateToProps and pass the user into this function to link it to the store.

assets/js/components/audit_trails/AuditTable.jsx
```diff
@@ -20,123 +23,47 @@ import TableFooterPagination from '../common/TableFooterPagination'
...
+      },
+      {
+        Header: 'Action',
+        accessor: 'action',
+        Cell: props => <span>{props.row.action}</span>
+      },
+      {
+        Header: 'Description',
+        accessor: 'description',
+        Cell: props => <span>{props.row.description}</span>
```
allenan: for cases where you're just tossing the value in a `<span>`, you can just specify the `accessor:` and leave off the `Cell:`

assets/js/components/teams/TeamShow.jsx
allenan: what if we defined 2 components, `<UserCan ...>` and `<UserCannot ...>`? Then you could have one thing displayed in the `true` case and a different thing in the `false` case

assets/js/components/teams/MembersTable.jsx
```diff
@@ -1,55 +1,81 @@
...
+          {userCan('update', 'membership', props.row) &&
+            <Button
+              onClick={() => openEditMembershipModal(props.row)}
+              size="small"
+            >
+              Edit
+            </Button>
+          }
+
+          {userCan('delete', 'membership', props.row) &&
```
allenan: we should be able to use the regular `<UserCan>` component here

assets/js/components/teams/TeamShow.jsx
```diff
@@ -130,38 +113,49 @@ class TeamShow extends Component {
...
+      path: "/teams/activity",
+    }]
 
-  return {
-    currentTeamId,
-    memberships,
-    invitations
+    return (
+      <div>
+        <UserCan type="not" action="view" itemType="auditTrails">
```
vicmgs: Added UserCan with type of "not". Let me know if you like a separate component instead.

Also, with this setup, there's a visual bug when you refresh the browser on /teams/activity. It flashes the members and invitations tables briefly due user being null in the beginning. Ive tried many things and think we must have the user injected into the component through mapStateToProps, which allows us to look for when the user.role == null to prevent the flashing of the member and invite tables.

assets/js/components/common/UserCan.jsx
```diff
@@ -23,7 +23,8 @@ class UserCan extends Component {
   }
 
   render() {
-    if (this.userCan()) return this.props.children
+    if (this.props.type == "not" && !this.userCan()) return this.props.children
```
allenan: i'd recommend breaking it into 2 component, `<UserCan>` and `<UserCannot>` because it'll be more apparent when looking through code what's going on, and also, we can just have separate render methods with their own logic instead of switching on `props.type` here

assets/js/components/teams/TeamShow.jsx
allenan: we have access to the user in the `<UserCan>` component right? can we just prevent rendering if role is null?

assets/js/components/common/UserCan.jsx
allenan: trick will be making sure not to duplicate the rules logic

vicmgs:
sounds good

PR #128, opened 2018-06-12
Team switching fix, remove useless code

vicmgs:
This removes useless code on the front end, still need to remove the now unnecessary socket stuff on the backend.


vicmgs:
ok just to clarify, graphql would be the call that returns the access error, so I would listen for the access error on the DeviceShow component and redirect from there right?

assets/js/util/apolloClient.js
```diff
@@ -24,7 +26,25 @@ const authLink = setContext((_, { headers }) => {
   }
 })
 
-const authHttpLink = authLink.concat(httpLink)
+const authErrorLink = onError(({ networkError, operation: { operationName }}) => {
+  if (networkError.statusCode == 404) {
```
vicmgs: not sure if listening for 404s is correct here, or if I should make it so that the device/channel/gateway show query should return no errors but return a successful response with a message that says none was found. Let me know what you think

PR #129, opened 2018-06-13
Cleanup + add user two factor status to membership table

vicmgs:


lib/console/teams/membership_resolver.ex
```diff
@@ -6,10 +6,12 @@ defmodule Console.Teams.MembershipResolver do
       Ecto.assoc(current_team, [:memberships])
       |> Repo.paginate(page: page, page_size: page_size)
 
-    entries = memberships.entries |> Repo.preload(:user)
+    entries = memberships.entries |> Repo.preload([user: [:twofactor]])
```
allenan: this works. if you want to be fancy, you could do something like how notifications gets its virtual `active` column. might be a little more efficient. your call though

assets/js/components/common/SmallChip.jsx
```diff
@@ -3,9 +3,14 @@ import Chip from '@material-ui/core/Chip';
 import { withStyles } from '@material-ui/core/styles'
 
 const styles = theme => ({
-  chip: {
```
allenan: enabled/disabled two factor seems out of scope for this component (SmallChip)

lib/console/teams/membership_resolver.ex
allenan: it would be 1 query instead of 3 (I think) and you wouldn't need to load all the associated fields for users and twofactors into memory

lib/console/teams/membership.ex
```diff
@@ -28,4 +33,12 @@ defmodule Console.Teams.Membership do
     membership
     |> changeset(%{user_id: user.id, team_id: team.id, role: role})
   end
+
+  def user_twofactor(query) do
+    from u in User,
+    inner_join: m in ^query, on: [user_id: u.id],
+    left_join: t in TwoFactor, on: [user_id: u.id],
+    where: not is_nil(u.id),
+    select: %{m | email: u.email, two_factor_enabled: not is_nil(t.id)}
```
allenan: nice!

vicmgs: your original was very helpful 👍

PR #130, opened 2018-06-13
store private/public key and address on teams

allenan:
this is a prereq for gateway registration

- integrate libp2p as dep for crypto functions
- when creating teams, assign them a private key and corresponding public key & address
- teams will be the owner/payee address for gateways (and later devices)


lib/console/teams/team.ex
```diff
@@ -6,15 +6,21 @@ defmodule Console.Teams.Team do
...
-    has_many :invitations, Console.Teams.Invitation
+    has_many :invitations, Console.Teams.Invitation, on_delete: :delete_all
 
-    has_many :devices, Console.Devices.Device
-    has_many :gateways, Console.Gateways.Gateway
-    has_many :channels, Console.Channels.Channel
-    has_many :audit_trails, Console.AuditTrails.AuditTrail
-    has_many :notifications, Console.Notifications.Notification
+    has_many :devices, Console.Devices.Device, on_delete: :delete_all
+    has_many :gateways, Console.Gateways.Gateway, on_delete: :delete_all
```
vicmgs: does this mean the gateways get deleted if the team gets deleted? isn't that bad because a gateway could be owned by someone not in the team?

lib/console/teams/team.ex
```diff
@@ -6,15 +6,21 @@ defmodule Console.Teams.Team do
...
+    has_many :channels, Console.Channels.Channel, on_delete: :delete_all
+    has_many :audit_trails, Console.AuditTrails.AuditTrail, on_delete: :delete_all
```
vicmgs: Do we want to delete these? might be good to keep

mix.exs
```diff
@@ -52,7 +52,8 @@ defmodule Console.Mixfile do
       {:scrivener_ecto, "~> 1.0"},
       {:absinthe, "~> 1.4"},
       {:absinthe_plug, "~> 1.4"},
-      {:absinthe_phoenix, "~> 1.4"}
+      {:absinthe_phoenix, "~> 1.4"},
+      {:libp2p, github: "helium/erlang-libp2p", branch: "master"}
```
vicmgs: 💯

lib/console/teams/team.ex
allenan: in this case, we're not exposing team deletion to users yet. This was just so I could clean up my teams locally -- I had too many from testing.

but also, teams own gateways, users don't. that's why the team gets the private key. any actions team members take are on behalf of the team

lib/console/teams/team.ex
allenan: later, yeah. for now, I just wanted to be able to clean things up locally without triggering foreign key constraint errors

PR #131, opened 2018-06-14
Add initial front end for gateway registration

vicmgs:


vicmgs:
After some thinking, this is how I pictured the front end will update to reflect the gateway registration progress. Let me know if you agree

1. User selects a name for gateway and scans QR codes, the phone app sends the QR code info out through p2p network. The front end at this point joins a socket connection.
2. When router hits console with the initial signed payload, console creates the gateway object in the db, tells the front end to render loading status with the socket connection. 
3. When router publishes to blockchain successfully and tells console, console changes gateway status to verified, and tells the front end to render the success status with the socket connection.
4. User can now view the gateway.

One question is, what do we do if the user navigates away from the page during the loading progress? It doesn't seem to matter to me since most of the important work is done over the backend and the user can simply see the successful gateway in the gateway index once done. The issue is with if the registration fails, but we could just add a message on the front end to say "DO not navigate away from this page"

vicmgs:
screen shot 2018-06-14 at 4 04 38 pm
screen shot 2018-06-14 at 4 04 50 pm
screen shot 2018-06-14 at 4 05 08 pm
screen shot 2018-06-14 at 4 05 25 pm

vicmgs:
Made changes so that step 2 of gateway registration continues on gatewayShow component.

Currently, view only users can directly visit the gateways/new url. There is no issue with a view only user creating a gateway as the backend prevents it, but is there a good way to redirect view only users from hitting that url manually?

PR #133, opened 2018-06-18
Gateway registration router API

vicmgs:


lib/console_web/controllers/gateway_controller.ex
```diff
@@ -61,6 +61,34 @@ defmodule ConsoleWeb.GatewayController do
...
+    #   gateway["payee_address"] is the team b58b58_address
+    #
+    # First check the following is valid:
+    #   nonce verification is valid and not messed with
+    #   OUI matches console's oui
+    #
+    # Then update gateway's public key
+    #
+    # If nonce and OUI verification fails, send back error status 404
+    # If verification succeeds, send back countersigned data
```
vicmgs: How do I countersign the data? Is it part of libp2p?

lib/console_web/controllers/gateway_controller.ex
```diff
@@ -61,6 +61,34 @@ defmodule ConsoleWeb.GatewayController do
     end
   end
 
+  def register(conn, %{"OUI" => oui, "nonce" => nonce, "gateway" => gateway}) do
+    # receive the following from router:
+    #   nonce and OUI self explanatory
+    #   gateway["id"] is 32 bit gateway ID
```
vicmgs: Should we randomly generate a 32 bit random gateway id? 

Do we convert the 32 bit binary into a list and send it over http? So what router would send back is a list that we can use list_to_binary on

lib/console_web/controllers/gateway_controller.ex
allenan: i think that part is tdb. it'll be part of the transaction system eventually, but Rahul is still working on that. so you can just put some placeholder in there for now

lib/console_web/controllers/gateway_controller.ex
allenan: i'd imagine we'd pass it as a string since that's a little more compatible with json apis

allenan: and `gateway["id"]` is the ID in OUI/ID right? not the gateway row's primary key id?

lib/console_web/controllers/gateway_controller.ex
```diff
@@ -61,6 +61,34 @@ defmodule ConsoleWeb.GatewayController do
     end
   end
 
+  def register(conn, %{"OUI" => oui, "nonce" => nonce, "gateway" => gateway}) do
```
allenan: you probably want to pattern match into the gateway as well, so that it's super clear what's coming through there

vicmgs: sounds good

lib/console_web/controllers/gateway_controller.ex
vicmgs: yup, have yet to make that table, was figuring out how to generate the binary first. 

Will probably use :crypto.strong_rand_bytes(4)

lib/console_web/router.ex
```diff
@@ -66,6 +66,8 @@ defmodule ConsoleWeb.Router do
     pipe_through ConsoleWeb.RouterApiPipeline
 
     resources "/devices", DeviceController, only: [:show]
+    post "/gateway/register", GatewayController, :register
```
allenan: I believe convention here would be for `gateways` to be plural

lib/console/gateways/gateways.ex
```diff
@@ -105,4 +111,25 @@ defmodule Console.Gateways do
...
+          gatewayIdentifier = %GatewayIdentifier{} -> gatewayIdentifier |> Repo.preload(:gateway)
+          nil -> :error
+        end
+      :error -> :error
+    end
+  end
+
+  defp create_gateway_identifier do
+    attrs = %{
+      unique_identifier: :crypto.strong_rand_bytes(4)
```
vicmgs: do we need to worry about handling the case this random function generates the same identifier?

lib/console_web/router.ex
```diff
@@ -66,6 +66,8 @@ defmodule ConsoleWeb.Router do
     pipe_through ConsoleWeb.RouterApiPipeline
```
vicmgs: Why is the scope here under ConsoleWeb.Router and not just ConsoleWeb?

priv/repo/migrations/20180618202035_create_gateway_identifiers_table.exs
```diff
@@ -0,0 +1,17 @@
+defmodule Console.Repo.Migrations.CreateGatewayIdentifiersTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:gateway_identifiers, primary_key: false) do
```
allenan: so these won't just be for gateways. We'll want to be able to apply these identifiers to any hardware our OUI is responsible for and keep them unique across all of those things. right now that's devices and gateways but we want to keep it flexible

lib/console_web/views/gateway_view.ex
```diff
@@ -23,6 +23,13 @@ defmodule ConsoleWeb.GatewayView do
     |> append_events(gateway.events)
   end
 
+  def render("gateway_register.json", %{tx: tx, signature: signature}) do
```
allenan: this should probably be within the `views/router/` scope in order to keep this separate from any user-facing API

lib/console_web/router.ex
allenan: because router has a different authentication scheme

allenan: and can access data and routes that a regular user cannot

lib/console_web/controllers/gateway_controller.ex
```diff
@@ -61,6 +62,29 @@ defmodule ConsoleWeb.GatewayController do
     end
   end
 
+  def register(conn, %{"OUI" => oui, "nonce" => nonce, "gateway" => %{"id" => gateway_id, "public_key" => public_key, "payee_address" => payee_address}}) do
```
allenan: this should be within the router gateway controller, not the user-facing one

lib/console/gateways/gateways.ex
allenan: it will raise a database unique constraint error. we should be able to rescue from that and retry n number of times (it'll be a really low probability of happening so a couple retries will be way more than enough)

lib/console/gateways/gateways.ex
```diff
@@ -105,4 +111,25 @@ defmodule Console.Gateways do
...
+      {:ok, identifier} ->
+        case Repo.get_by(GatewayIdentifier, unique_identifier: identifier) do
+          gatewayIdentifier = %GatewayIdentifier{} -> gatewayIdentifier |> Repo.preload(:gateway)
+          nil -> :error
+        end
+      :error -> :error
+    end
+  end
+
+  defp create_gateway_identifier do
```
allenan: given the comment below on the db table, we probably want to put this in its own context, maybe `HardwareIdentifiers` or something?

lib/console/gateways/gateways.ex
```diff
@@ -90,7 +95,8 @@ defmodule Console.Gateways do
 
   """
   def delete_gateway(%Gateway{} = gateway) do
-    Repo.delete(gateway)
+    Repo.get!(GatewayIdentifier, gateway.gateway_identifier_id)
```
allenan: if you add `on_delete: :delete_all` to the relationship on the gateway side, you should be able to just  delete the gateway and that will cascade to the hardware identifiers

lib/console/gateways/gateway_identifier.ex
```diff
@@ -0,0 +1,22 @@
...
+
+    has_one :gateway, Gateway, on_delete: :delete_all
+
+    timestamps()
+  end
+
+  def changeset(gateway_identifier, attrs) do
+    gateway_identifier
+    |> cast(attrs, [:unique_identifier])
+    |> validate_required([:unique_identifier])
```
allenan: i think you can also add a unique validation here and it will return a handleable error instead of a raw database constraint violation crash

lib/console/gateways/gateway_identifier.ex
```diff
@@ -0,0 +1,22 @@
+defmodule Console.Gateways.GatewayIdentifier do
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  alias Console.Gateways.Gateway
+
+  @primary_key {:id, :binary_id, autogenerate: true}
+  @foreign_key_type :binary_id
+  schema "gateway_identifiers" do
+    field :unique_identifier, :binary
```
allenan: i might just call this `token` or something for conciseness

lib/console_web/router.ex
vicmgs: Is there a separate file for ConsoleWeb.Router? I get the RouterApiPipeline is different than AuthApiPipeline but dont' get why I can't hit the router through postman when scope "/api/router" is under ConsoleWeb.Router and not ConsoleWeb

lib/console_web/controllers/gateway_controller.ex
vicmgs: wow I'm blind... thanks

lib/console_web/router.ex
vicmgs: nvm lol i figured it out with your later comments

lib/console/gateways/gateway.ex
```diff
@@ -18,16 +19,16 @@ defmodule Console.Gateways.Gateway do
     field :status, :string
 
     belongs_to :team, Team
+    belongs_to :hardware_identifier, HardwareIdentifier
```
vicmgs: tried to put on delete: delete all here, would not compile =[

priv/repo/migrations/20180618202035_create_hardware_identifiers_table.exs
```diff
@@ -0,0 +1,17 @@
...
+    create table(:hardware_identifiers, primary_key: false) do
+      add :id, :binary_id, primary_key: true
+      add :token, :binary, null: false
+      timestamps()
+    end
+
+    create unique_index(:hardware_identifiers, [:token])
+
+    alter table(:gateways) do
+      add :hardware_identifier_id, references(:hardware_identifiers, on_delete: :nothing, type: :binary_id), null: false
```
allenan: does `on_delete: :nothing` make sense here?

vicmgs: made the fix. Also it seems the on_delete: :delete_all in hardware_identifier.ex does the same thing.

lib/console/gateways/gateways.ex
```diff
@@ -53,10 +54,36 @@ defmodule Console.Gateways do
       {:error, %Ecto.Changeset{}}
 
   """
-  def create_gateway(attrs \\ %{}) do
-    %Gateway{}
-    |> Gateway.changeset(attrs)
-    |> Repo.insert()
+  def create_gateway(attrs \\ %{}, attempt \\ 1) do
+    token = :crypto.strong_rand_bytes(4)
```
allenan: this token generation (and also maybe the retry logic?) probably belongs in the hardware identifier schema file

lib/console/gateways/gateway.ex
allenan: ok

lib/console/hardware_identifiers/hardware_identifiers.ex
```diff
@@ -0,0 +1,17 @@
...
+  import Ecto.Query, warn: false
+  alias Console.Repo
+
+  alias Console.HardwareIdentifiers.HardwareIdentifier
+
+  def get_resource_by_hardware_identifier(token, resource) do
+    case Base.decode32(token) do
+      {:ok, token} ->
+        case Repo.get_by(HardwareIdentifier, token: token) do
+          hardwareIdentifier = %HardwareIdentifier{} -> hardwareIdentifier |> Repo.preload(resource)
```
allenan: i think you could probably do this with a single query (there are 2 trips to the db/Repo in this function)

lib/console/gateways/gateways.ex
```diff
@@ -53,10 +54,35 @@ defmodule Console.Gateways do
       {:error, %Ecto.Changeset{}}
 
   """
-  def create_gateway(attrs \\ %{}) do
-    %Gateway{}
-    |> Gateway.changeset(attrs)
-    |> Repo.insert()
+  def create_gateway(attrs \\ %{}, attempt \\ 1) do
```
vicmgs: moved the token generation to HardwareIdentifier.ex 

The reattempts can't be moved because we are using ecto multi, so the unique token failure doesn't trigger until Ecto.Multi.run in this file is reached.

lib/console/hardware_identifiers/hardware_identifier.ex
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def get_associated_resource(token, resource) do
+    table =
+      case resource do
+        :gateway -> Gateway
+      end
+
+    query = from h in HardwareIdentifier,
+      left_join: t in ^table, on: [hardware_identifier_id: h.id],
```
allenan: if you passed in a straight ecto schema like `Gateway` for resource, rather than `:gateway`, could you use that directly in place of `^table`? (to avoid the case)

lib/console/gateways/gateways.ex
allenan: ok that's fine

lib/console/hardware_identifiers/hardware_identifier.ex
vicmgs: niceeeeee

allenan: also the statement reads a little backwards. you might want to start with `query = from r in resource` and then join to `HardwareIdentifier`. then you don't have to swap out the `select` at the end. usually you'd start with the thing you want to end up with

PR #134, opened 2018-06-20
Set up socket connections for gateway registration

vicmgs:


assets/js/components/SocketHandler.jsx
```diff
@@ -12,64 +11,35 @@ class SocketHandler extends Component {
...
-      .receive("ok", resp => { console.log(`Joined ${channelName} successfully`, resp) })
-      .receive("error", resp => { console.log(`Unable to join ${channelName}`, resp) })
-
-    channel.on("new", res => newHandler(res))
-    channel.on("delete", res => deleteHandler(res))
-
-    if (updateHandler !== undefined) {
-      channel.on("update", res => updateHandler(res))
+      console.log('connecting')
+      // join goes here
```
allenan: do we need to add the join here?

assets/js/components/SocketHandler.jsx
```diff
@@ -12,64 +11,35 @@ class SocketHandler extends Component {
...
     this.disconnect = this.disconnect.bind(this)
-    this.join = this.join.bind(this)
   }
 
-  componentDidMount() {
-    if (this.props.isLoggedIn && this.props.currentTeamId) {
-      this.props.fetchIndices()
-      this.subscribeToUpdates()
-    }
+  componentWillMount() {
```
allenan: if we changed this to `<UserTeamProvider />`, what are we still using this component for?

assets/js/components/SocketHandler.jsx
```diff
@@ -83,24 +53,21 @@ class SocketHandler extends Component {
   render() {
     return (
       <div>
-        {this.props.children}
+        {React.cloneElement(this.props.children, { socket: this.socket, currentTeamId: this.props.currentTeamId })}
```
allenan: how are these being accessed?

allenan: also have you looked into the new context api? https://reactjs.org/docs/context.html

assets/js/components/gateways/GatewayRegister.jsx
```diff
@@ -32,9 +32,27 @@ class GatewayRegister extends Component {
...
+    const channelName = `gateway:${gateway.id}`
+
+    let channel = socket.channel(channelName, {team_id: currentTeamId})
+
+    channel.join()
+      .receive("ok", resp => { console.log(`Joined ${channelName} successfully`, resp) })
+      .receive("error", resp => { console.log(`Unable to join ${channelName}`, resp) })
+
+    channel.on("registering", () => { this.setState({showQRCode: false}) })
+    channel.on("verified", () => { window.location.reload() })
```
allenan: why do we need to force a reload when the status changes to verified?

assets/js/components/gateways/GatewayShow.jsx
```diff
@@ -68,7 +71,16 @@ class GatewayShow extends Component {
 
     if (gateway.status == "pending") return (
       <DashboardLayout title={gateway.name}>
-        <GatewayRegister gateway={gateway} />
+        <UserCan>
```
allenan: what does it mean to use `<UserCan>` and cannot without any arguments?

lib/console_web/controllers/router/gateway_controller.ex
```diff
@@ -23,11 +24,16 @@ defmodule ConsoleWeb.Router.GatewayController do
...
 
         conn
         |> send_resp(:no_content, "")
     else _ ->
       {:error, :not_found}
     end
   end
+
+  defp broadcast(%Gateway{} = gateway, action) do
+    ConsoleWeb.Endpoint.broadcast("gateway:#{gateway.id}", action, %{})
```
allenan: do we want to use the plain phoenix channels over graphql subscriptions?

vicmgs: hmmm let me think about this one, didn't think to use it since we aren't fetching anything with graphql for GatewayRegister

assets/js/components/gateways/GatewayRegister.jsx
vicmgs: this allows gateway show page to repull the graphql data, otherwise, it doesn't show the verified gateway and stay on the loading status

assets/js/components/SocketHandler.jsx
vicmgs: I envisioned this component as a wrapper for components that need socket access, gonna look into the graphql subscriptions tho

PR #135, opened 2018-10-30
Upgrade elixir version to 1.7.4, travis otp 21.1

vicmgs:


allenan:
should we go with 1.7.4 since that's the current version?

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: elixir
 elixir:
-  - '1.6.2'
+  - '1.7.3'
 otp_release:
   - '20.2.4'
```
allenan: also probably want otp 21.1

PR #405, opened 2021-03-05
Update device stats with new event log, add DC used

vicmgs:


lib/console_web/controllers/router/device_controller.ex
```diff
@@ -172,19 +173,35 @@ defmodule ConsoleWeb.Router.DeviceController do
...
                 false -> 0
               end
+            packet_count = if dc_used == 0, do: 0, else: 1
+
             Devices.update_device(device, %{
               "last_connected" => event.reported_at_naive,
               "frame_up" => event.data["frame_up"],
               "frame_down" => event.data["frame_down"],
-              "total_packets" => device.total_packets + 1,
+              "total_packets" => device.total_packets + packet_count,
```
vicmgs, line 187: I think we only want to count uplinks, im like 95% sure lol

ambuznego: I think you right

ambuznego (approved):
:shipit:

PR #418, opened 2021-03-12
Deleting device on device show page does not remove from device table upon redirect

ambuznego:
PR ensures we refetch after devices is done loading upon redirect from deleting in device show page

```
...............................................

Finished in 1.1 seconds
49 tests, 0 failures
```


vicmgs (approved):
nice find

PR #422, opened 2021-03-16
Fix debug - integration always showing

ambuznego:
Should be using state to check whether integration info should be showing or not (based on the eye button)

### DIFF
--- a/assets/js/components/common/DebugEntry.jsx
+++ b/assets/js/components/common/DebugEntry.jsx
@@ -204,7 +204,7 @@ class DebugEntry extends Component {
                       "id": event.data.integration.id,
                       "name": event.data.integration.name,
                       "status": event.data.integration.status,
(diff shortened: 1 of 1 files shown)


assets/js/components/common/DebugEntry.jsx
```diff
@@ -196,7 +196,7 @@ class DebugEntry extends Component {
             )
           }
           {
-            event.data.integration && (
+            this.state.showIntegrationInfo && (
```
vicmgs, line 199: you can't show integration info if event.data.integration doesnt exist right?

vicmgs (approved):

assets/js/components/common/DebugEntry.jsx
ambuznego: you right, otherwise it'd crap out again, added condition

PR #423, opened 2021-03-16
Fix events to display 0 as fct for joins

ambuznego:


assets/js/components/events/EventsDashboard.jsx
```diff
@@ -268,7 +269,7 @@ class EventsDashboard extends Component {
         reported_at: firstEvent.reported_at,
         category: firstEvent.category,
         sub_categories: orderedRouterEvents.map(e => e.sub_category),
-        fct: firstEvent.frame_up || firstEvent.frame_down,
+        fct: firstEvent.frame_down || firstEvent.frame_up,
```
ambuznego, line 272: since `frame_up` is being used for all joins (when it is 0), the order is important to make sure it doesn't come back as null

vicmgs (approved):

PR #450, opened 2021-03-19
Keep scroll position in debug panel

ambuznego:


assets/js/components/common/Debug.jsx
```diff
@@ -33,6 +38,21 @@ class Debug extends Component {
     }
   }
 
+  getSnapshotBeforeUpdate(prevProps, prevState) {
```
vicmgs, line 41: does this function get called? how does this work

ambuznego: Gets called before committing to the DOM and its result comes into `componentDidUpdate` through `snapshot`

https://reactjs.org/docs/react-component.html#getsnapshotbeforeupdate

vicmgs: magic

vicmgs (approved):

PR #457, opened 2021-03-22
Event log JSON export all events

ambuznego:


lib/console_web/router.ex
```diff
@@ -40,6 +40,7 @@ defmodule ConsoleWeb.Router do
     post "/devices/delete", DeviceController, :delete
     post "/devices/debug", DeviceController, :debug
     post "/devices/set_active", DeviceController, :set_active
+    get "/events/:device_id", DeviceController, :get_all_events
```
vicmgs, line 43: maybe change this to /devices/:device_id/events?

vicmgs (approved):
just 1 thing, nice!

lib/console_web/router.ex
ambuznego: good point, yeshh

PR #474, opened 2021-03-25
Drop last event when adding new event

ambuznego:
Preventing the orphaning of events when dropping the oldest event at the time of adding a new one so that the event log doesn't get annoyingly long on the browser. Leaving splice on render as a way to ensure what we grab from db isn't orphaning as well.


vicmgs (approved):
may god have mercy on you

PR #476, opened 2021-03-26
Discovery Mode

ambuznego:
- Add `/v1/devices/discover` endpoint for Wallet API to use
- Prevent organization creation with Discovery Mode org name
- Prevent creating a device with `hotspot_address` from other endpoints
- Remove device creation limit for Discovery Mode org
- Bypass the `rate_limit` check for too many requests in `test` environment

Tests passing:
```
.................................................

Finished in 1.1 seconds
51 tests, 0 failures
```

### DIFF
--- a/lib/console/devices/devices.ex
+++ b/lib/console/devices/devices.ex
@@ -70,6 +70,12 @@ defmodule Console.Devices do
     Repo.all(query)
   end
 
+  def get_device_for_hotspot_address(address) do
+    Device
+      |> where([d], d.hotspot_address == ^address)
+      |> Repo.one()
+  end
+
   def fetch_assoc(%Device{} = device, assoc \\ [:events, :organization, :channels]) do
     Repo.preload(device, assoc)
   end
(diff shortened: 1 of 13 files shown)


lib/console/devices/devices.ex
```diff
@@ -70,19 +70,32 @@ defmodule Console.Devices do
...
     cond do
-      count > 9999 ->
+      organization.name !== "Discovery Mode (Helium)" and count > 9999 ->
         {:error, :forbidden, "Device limit for organization reached"}
       true ->
-        %Device{}
-        |> Device.create_changeset(attrs)
-        |> Repo.insert()
+        cond do
+          attrs["hotspot_address"] != nil ->
```
vicmgs, line 90: in the regular create_device controller action, we need to drop the hotspot_address attr because hackers could manually add it in and hijack through certain api tools

vicmgs: same in the v1 device controller

ambuznego: do we need to in `v1` since the params specified and thus it wouldn't match to the `create` function and fail?

ambuznego: I'll test though :-)

vicmgs: you are correct, v1 does not need!

ambuznego: addressed 👍 `v1` not needed per se but I added just to be safe.

vicmgs (approved):
nice one!

vicmgs (approved):

PR #487, opened 2021-03-31
Add device index label bar 

vicmgs:


ambuznego:
adding a device to the label does not update # of devices on the label bar
image

ambuznego (approved):

PR #490, opened 2021-04-01
Update individual label show pages on deviceIndex labels bar

vicmgs:
removes label routes and components that are unused as well


ambuznego:
after I remove a label from a device and I go to the label table, then it still shows the device. though it might be one of those weird caching things when hot reloading?

ambuznego:
clicking the device name does not take me to the device's show page when on a label's table

ambuznego (approved):
v nice

PR #501, opened 2021-04-06
Show no integration attached when no_channel in external integrations

ambuznego:


assets/js/components/events/EventsDashboard.jsx
```diff
@@ -359,7 +359,10 @@ class EventsDashboard extends Component {
         render: (data, row) => {
           const integrationResponses = row.integrations && row.integrations.filter(i => i.subCategory === 'uplink_integration_res');
           const integrationError = integrationResponses.findIndex(i => i.status === 'error') === -1 ? false : true;
-          const integrationMissing = row.integrations.findIndex(i => i.id === 'no_channel') === -1 ? false : true;
+
+          const externalIntegrations = row.integrations.filter(i => i.name !== 'Internal Integration');
+          const integrationMissing = externalIntegrations.findIndex(i => i.id === 'no_channel') === -1 ? false : true;
```
vicmgs, line 364: might need to add a case where if externalIntegrations length is 0, it should be true

ambuznego: true! addressed.

assets/js/components/events/EventsDashboard.jsx
```diff
@@ -360,9 +360,11 @@ class EventsDashboard extends Component {
           const integrationResponses = row.integrations && row.integrations.filter(i => i.subCategory === 'uplink_integration_res');
           const integrationError = integrationResponses.findIndex(i => i.status === 'error') === -1 ? false : true;
 
-          const externalIntegrations = row.integrations.filter(i => i.name !== 'Internal Integration');
-          const integrationMissing = externalIntegrations.findIndex(i => i.id === 'no_channel') === -1 ? false : true;
-          
+          const externalIntegrations = row.integrations && row.integrations.filter(i => i.name !== 'Internal Integration');
```
vicmgs, line 363: do you want to default externalIntegrations to [] here?

otherwise you would have to check for its existence in the next 2 lines

ambuznego: good thinking, it worked ok for when the event didn't have integration, but best to be safe.

vicmgs (approved):
senddddd

PR #517, opened 2021-04-14
Node handles on flows drag and drop sidebar 

vicmgs:


ambuznego (approved):
LGTM, only thing I noticed was that the nodes coming from sidebar have a little bit of white background in the handle area but not sure we can get around that

PR #518, opened 2021-04-15
Device flows component

ambuznego:
image

Left `TODO` item regarding the size calculation of the nodes


lib/console/flows/flow_resolver.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Console.Flows.FlowResolver do
+  alias Console.Repo
+  alias Console.Flows.Flow
+  alias Console.Labels
+  import Ecto.Query
+
+  def get_by_device(%{device_id: device_id}, %{context: %{current_organization: current_organization}}) do
+    labels = Labels.get_device_labels(device_id)
+    label_ids = Enum.map(labels, fn (l) -> l.label_id end)
+    IO.inspect label_ids
```
vicmgs, line 10: log

assets/js/components/devices/DeviceFlows.jsx
```diff
@@ -0,0 +1,193 @@
...
+      while (existingElements[nodeId + "_copy" + i]) {
+        i++;
+      }
+      return `${nodeId}_copy${i}`;
+    } else {
+      return nodeId;
+    }
+  }
+
+  generateElements = (flows) => {
```
vicmgs, line 43: to clarify, all we are doing here is creating a row for each flow row that is returned from the db right? and making copies of nodes as needed

assets/js/components/common/nodes/ChannelNode.jsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
+
+function mapStateToProps(state) {
+  return {
+    socket: state.apollo.socket
+  };
+}
+
+export default connect(mapStateToProps, null)(
```
vicmgs, line 42: Can we instead of making numerous gql calls on the nodes themselves, make a single gql query for all the node ids once the deviceFlowsQuery has returned?

ambuznego: yep that's fair, I debated doing that, will change 👍

assets/js/components/devices/DeviceFlows.jsx
ambuznego: correcto!

lib/console/flows/flow_resolver.ex
ambuznego: oopsie

assets/js/components/common/nodes/ChannelNode.jsx
vicmgs: tyty

vicmgs (approved):
woop woop

PR #523, opened 2021-04-16
Device flows fixes

ambuznego:
- Fixes layout which was sometimes not rendering the nodes or rendering them in a weird order (theory is because the graph wasn't recreated on re-render)
- Shows text when no flows exist
- Fixes the device name not showing up when it's part of a flow w/ only label


vicmgs (approved):
🥇

PR #533, opened 2021-04-19
Prevent read-only users from seeing hidden attributes

vicmgs:
channel downlink_token hidden
device app_key hidden
organization webhook_key hidden


ambuznego (approved):
👍

PR #534, opened 2021-04-19
Import device move to device new screen

vicmgs:


assets/js/components/devices/import/ChooseImportType.jsx
```diff
@@ -48,8 +49,9 @@ const ChooseImportType = ({onImportSelect, scanGenericDevices, genericImportScan
   )
   return (
     <Fragment>
-      <Title style={{fontSize: 40}}>Import Devices</Title>
-      <Text style={{width: '100%', textAlign: 'center', margin: '0px 40px 20px'}}>
+      <img src={ImportIcon} style={{marginBottom: 10, height: 50, objectFit: 'cover'}}/>
+      <Title style={{fontSize: 22}}>Import Devices</Title>
+      <Text style={{width: '100%', textAlign: 'center', margin: '0px 40px 10px'}}>
```
ambuznego, line 54: I'd increase font size for this bad boy

ambuznego:
can be separate ticket but probs makes sense to remove LabelId as an import column since user will no longer have this info available

ambuznego (approved):
minor suggestion + import stuck (though successful @ creating device) and does not show success

otherwise lgtm

PR #537, opened 2021-04-20
Properly format event body in debug panel

ambuznego:
- properly format body before creating event
- remove old test that is no longer relevant (no more `debug` endpoint)

image

passing tests
```
..................................................

Finished in 1.9 seconds
50 tests, 0 failures
```

### DIFF
--- a/lib/console_web/controllers/router/device_controller.ex
+++ b/lib/console_web/controllers/router/device_controller.ex
@@ -139,6 +139,13 @@ defmodule ConsoleWeb.Router.DeviceController do
         result =
           Ecto.Multi.new()
           |> Ecto.Multi.run(:event, fn _repo, _ ->
+            event = case event["data"]["req"]["body"] do
+              nil -> event
+              _ -> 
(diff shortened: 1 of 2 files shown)


lib/console_web/controllers/router/device_controller.ex
```diff
@@ -139,6 +139,13 @@ defmodule ConsoleWeb.Router.DeviceController do
         result =
           Ecto.Multi.new()
           |> Ecto.Multi.run(:event, fn _repo, _ ->
+            event = case event["data"]["req"]["body"] do
+              nil -> event
+              _ -> 
+                {:ok, decoded_body} = Poison.decode(event["data"]["req"]["body"])
+                Kernel.put_in(event["data"]["req"]["body"], decoded_body)
```
vicmgs, line 146: can we use Map.put instead?

ambuznego: can't do deeply nested using `Map.put` :/

vicmgs: ic ic

vicmgs (approved):

PR #540, opened 2021-04-21
Force logout when session token expires

vicmgs:
currently, local auth0 sessions will expire in 30 seconds so you can test this code on local for now. I'll revert the timer back to 10 hours once PR is merged.

### DIFF
--- a/assets/js/actions/apollo.js
+++ b/assets/js/actions/apollo.js
@@ -4,6 +4,7 @@ import { onError } from "@apollo/client/link/error";
 import { setContext } from "@apollo/client/link/context";
 import { store } from '../store/configureStore';
 import { replace } from 'connected-react-router';
+import { logout } from '../components/auth/Auth0Provider'
 import createSocket from '../socket'
 
 export const CREATED_APOLLO_CLIENT='CREATED_APOLLO_CLIENT';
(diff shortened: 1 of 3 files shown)


assets/js/actions/apollo.js
```diff
@@ -102,27 +103,20 @@ export const setupApolloClient = (getAuthToken, organizationId) => {
     socket.connect()
 
     store.subscribe(async () => {
-      const newTokenClaims = await getAuthToken();
-      const newToken = newTokenClaims.__raw;
-      const newOrganization = store.getState().organization;
-
-      if (newToken !== token || currentOrganizationId !== newOrganization.currentOrganizationId) {
```
vicmgs, line 109: none of this ever does anything, switch orgs reloads the page entirely, bypassing this

assets/js/util/rest.js
```diff
@@ -51,9 +50,15 @@ const headers = async () => {
   if (organizationId) {
     Object.assign(headerParams, { organization: organizationId })
   }
-  const tokenClaims = await getIdTokenClaims();
```
vicmgs, line 54: this generate a new token everytime, and thus never expires. This is why in the future state, I only use it for the first rest api call and use the stored token in global state once apolloClient has been created so we can keep using the same token.

assets/js/actions/apollo.js
```diff
@@ -102,27 +103,20 @@ export const setupApolloClient = (getAuthToken, organizationId) => {
...
-        currentOrganizationId = newOrganizationId
-        token = newToken;
-        socket.disconnect();
-        socket = createSocket(token, currentOrganizationId)
-        socket.connect()
+      if (Math.ceil(Date.now() / 1000) > store.getState().apollo.tokenClaims.exp) {
+        logout()
```
ambuznego, line 107: could we alert user before this happens and give them an opportunity to renew the session?

vicmgs: i still havent been able to figure out token refresh, so probably try to make that in a separate PR. In the meantime, I'll make the token expiration on prod like 36 hours or something

ambuznego:
seems like the session still expires if user is active, is that something we can address w/ auth0 to extend the session?

assets/js/actions/apollo.js
vicmgs: the docs are terrible

ambuznego: for sure, we can address the immediate urgency of logging user out and tackle refreshing session in future ticket to improve UX

ambuznego (approved):

PR #541, opened 2021-04-21
Refresh token if user acts in the last hour before token expires

vicmgs:


assets/js/actions/apollo.js
```diff
@@ -106,12 +60,80 @@ export const setupApolloClient = (getAuthToken, organizationId) => {
       if (Math.ceil(Date.now() / 1000) > store.getState().apollo.tokenClaims.exp) {
         logout()
       }
+
+      if (store.getState().apollo.tokenClaims.exp - Math.ceil(Date.now() / 1000) < 3600) {
```
vicmgs, line 64: can change 3600 to 35980 to for testing. Since token expires in 36000, you will have to wait 20 seconds to see the token refresh happen

ambuznego: is it supposed to be `3600` or `36000` in the code?

vicmgs: 36000 is the token expiration time. I put 3600 so if the user acts within the last hour of token expiration, it will reset

ambuznego: goootcha

assets/js/actions/apollo.js
```diff
@@ -106,12 +60,80 @@ export const setupApolloClient = (getAuthToken, organizationId) => {
...
+        const newAuthLink = setContext((_, { headers }) => {
+          return {
+            headers: {
+              ...headers,
+              authorization:`Bearer ${newTokenClaims.__raw}`,
+              organization: currentOrganizationId
+            }
+          }
+        })
+        const newLink = authErrorLink.concat(authLink.concat(httpLink))
```
ambuznego, line 75: is this supposed to be `newAuthLink`? `newAuthLink` is never used

vicmgs: Ooo yes ty

ambuznego (approved):

PR #583, opened 2021-05-04
Fix channel nodes not having integration icon

vicmgs:


assets/js/components/channels/ChannelBar.jsx
```diff
@@ -1,9 +1,35 @@
...
+  azure: AzureIcon,
+  cargo: CargoIcon,
+  cayenne: CayenneIcon,
+  datacake: DatacakeIcon,
+  http: HttpIcon,
+  mqtt: MqttIcon,
+  tago: TagoIcon,
+  ubidots: UbidotsIcon,
+  googlesheets: GoogleSheetIcon
+}
```
ambuznego, line 32: since we use he same object for `ChannelNode` too, maybe we can put this in a shared constant file and reuse for both?

ambuznego (approved):
just one tiny thing

assets/js/components/channels/ChannelBar.jsx
vicmgs: good point, ty!

PR #593, opened 2021-05-06
Provide helium ieee oui prefix suggestions for deveui and appeui

vicmgs:
https://app.clubhouse.io/hlm/story/9930/align-deveui-and-appeui-generation-with-lora-spec


ambuznego (approved):
texts need fixing, otherwise LGTM

PR #612, opened 2021-05-13
Inform router when devices and associated resources are updated

vicmgs:


lib/console/multi_buys/multi_buys.ex
```diff
@@ -25,4 +27,29 @@ defmodule Console.MultiBuys do
     |> MultiBuy.changeset(attrs)
     |> Repo.update()
   end
+
+  def get_all_multi_buy_associated_device_ids(id) do
+    device_ids =
+      get_associated_multi_buy_devices(id)
```
ambuznego, line 33: can we just do a select of ids so we don't have to map over the results?

vicmgs: good idea!

ambuznego: though if it gives you `[%{ id: 'someid' }, %{ id: 'someotherid' }]`, you may have to iterate anyway?

ambuznego: maybe  you could use `uniq_by` in that case?

lib/console/multi_buys/multi_buys.ex
```diff
@@ -25,4 +27,29 @@ defmodule Console.MultiBuys do
...
+
+    label_device_ids =
+      get_associated_multi_buy_labels(id)
```
ambuznego, line 36: probs same with this, using select

ambuznego (approved):
lgtm

PR #620, opened 2021-05-17
Update router get device info endpoint

vicmgs:


config/prod.exs
```diff
@@ -46,8 +46,7 @@ config :appsignal, :config,
   active: true,
   name: System.get_env("APPSIGNAL_APP_NAME"),
   push_api_key: System.get_env("APPSIGNAL_API_KEY"),
-  env: Mix.env,
-  ignore_actions: ["ConsoleWeb.Router.DeviceController#show"]
+  env: Mix.env
```
vicmgs, line 49: removing here since we want to keep track of this endpoint in the future in appsignal

lib/console/devices/devices.ex
```diff
@@ -48,7 +48,6 @@ defmodule Console.Devices do
 
   def get_by_dev_eui_app_eui(dev_eui, app_eui) do
      from(d in Device, where: d.dev_eui == ^dev_eui and d.app_eui == ^app_eui)
```
vicmgs, line 50: michael thinks fetching associated labels here is unnecessary

ambuznego (approved):

PR #621, opened 2021-05-17
Consistent input components

ambuznego:
To match the design for the alerts/multibuys name inputs and the backend rule of 50 chars limit, all NAME inputs should have limit on length and a length indicator 

NOTE:
besides linting, main changes are adding:
```
suffix={`${props.channelName.length}/50`}
maxLength={50}
```
and fixing styling for input components

image
image
image
image
image
image

...etc.


vicmgs (approved):
heck yea

PR #623, opened 2021-05-18
Fix permissions

ambuznego:


assets/js/components/flows/FlowsWorkspace.jsx
```diff
@@ -231,18 +238,22 @@ export default ({
             selectNodesOnDrag={false}
             multiSelectionKeyCode={null}
             selectionKeyCode={null}
+            nodesDraggable={userCan({ role: currentRole })}
```
vicmgs, line 241: are these still used?

vicmgs (approved):
Very nice! small question

assets/js/components/flows/FlowsWorkspace.jsx
ambuznego: this one, yes, since it prevents readonly user from dragging the nodes around, the 2nd one maybe redundant since I'm checking in the add+remove functions so I can take that one away

ambuznego: the 2nd one is also being used actually to prevent the edge from forming when you try to drag (even though it won't connect bc of the protection on the add function)
image

vicmgs: got it, thanks!

PR #624, opened 2021-05-19
Fix tests

ambuznego:
```
➜  console git:(fixTests) ✗ mix test
...................................................

Finished in 1.1 seconds
51 tests, 0 failures
```

### DIFF
--- a/lib/console_web/controllers/v1/device_controller.ex
+++ b/lib/console_web/controllers/v1/device_controller.ex
@@ -20,7 +20,7 @@ defmodule ConsoleWeb.V1.DeviceController do
       0 ->
         {:error, :not_found, "Device not found"}
       _ ->
-        render(conn, "show.json", device: List.first(devices))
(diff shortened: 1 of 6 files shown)


lib/console_web/controllers/v1/device_controller.ex
```diff
@@ -42,21 +42,21 @@ defmodule ConsoleWeb.V1.DeviceController do
...
 
-    render(conn, "index.json", devices: current_organization.devices)
+    render(conn, "index.json", devices: current_organization.devices |> Repo.preload([:labels]))
   end
 
   def show(conn, %{ "id" => id }) do
     current_organization = conn.assigns.current_organization
 
-    case Devices.get_device(current_organization, id) |> Devices.fetch_assoc([:labels]) do
+    case Devices.get_device(current_organization, id) |> Repo.preload([:labels]) do
```
ambuznego, line 59: since I could get `nil` and `fetch_assoc` cannot handle that

vicmgs (approved):

PR #625, opened 2021-05-19
Packet graph bug

ambuznego:
Currently, `no_channel` integration in the packet graph is being counted as an integration simply because it's part of the integrations array so it gets changed to blue color on the packet graph. It should not be counted as an integration in the event so it should remain red. This fix checks that the channel's `id` is NOT `no_channel`

### DIFF
--- a/assets/js/components/common/PacketGraph.jsx
+++ b/assets/js/components/common/PacketGraph.jsx
@@ -1,61 +1,65 @@
-import React, { Component } from 'react'
-import { Bubble } from 'react-chartjs-2'
+import React, { Component } from "react";
+import { Bubble } from "react-chartjs-2";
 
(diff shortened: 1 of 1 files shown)


assets/js/components/common/PacketGraph.jsx
```diff
@@ -108,7 +115,7 @@ class PacketGraph extends Component {
       const hotspots = event.hotspots;
 
       if (timeDiff < 300000) {
-        if (integrations.length > 0) {
+        if (integrations.length > 0 && integrations[0].id !== "no_channel") {
```
ambuznego, line 118: only change here, everything else is linting

vicmgs (approved):

PR #630, opened 2021-05-20
add alerts and multibuy to search

ambuznego:
image

image

image


assets/js/components/adr/AdrIndex.jsx
```diff
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
```
ambuznego, line 1: fixes bug introduced in last PR

vicmgs (approved):

PR #657, opened 2021-06-21
Clear selected nodes when remove selected node

ambuznego:
Previously, if you selected a node (that's not a function node), clicked on the button to remove the node, and then added it again through the node menu, it would show the selected icon because the internal state still considered it selected. This PR clears the internal selected state to prevent this. The `ReactFlowProvider` was moved to wrap the `FlowsWorkspace` component itself to prevent `react-redux context value` error.

### DIFF
--- a/assets/js/components/flows/FlowsWorkspace.jsx
+++ b/assets/js/components/flows/FlowsWorkspace.jsx
@@ -1,9 +1,9 @@
 import React, { useState, useRef, Fragment } from "react";
 import { useSelector } from "react-redux";
 import ReactFlow, {
-  ReactFlowProvider,
   isNode,
   isEdge,
+  useStoreActions,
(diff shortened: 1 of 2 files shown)


assets/js/components/flows/FlowsWorkspace.jsx
```diff
@@ -209,115 +212,114 @@ export default ({
...
 
-              setChangesState(true);
-              setShowInfoSidebar(false);
-            }}
-          />
-        </InfoSidebar>
-      </ReactFlowProvider>
+            setChangesState(true);
+            setShowInfoSidebar(false);
+            setSelectedElements([]);
```
ambuznego, line 319: main addition here, besides moving the provider to wrap `FlowsWorkspace` in the parent component

vicmgs (approved):

PR #664, opened 2021-06-24
Sending alert emails + webhooks async

ambuznego:


lib/console/jobs.ex
```diff
@@ -20,11 +20,15 @@ defmodule Console.Jobs do
...
     end)
 
     Enum.each(Enum.group_by(alertable_webhook_events, &Map.take(&1, [:alert_id, :event])), fn {identifiers, events} ->
-      send_webhook(identifiers, events)
+      Task.Supervisor.async_nolink(ConsoleWeb.TaskSupervisor, fn ->
+        send_webhook(identifiers, events)
+      end)
     end)
 
     Enum.each(alertable_email_events, fn e -> AlertEvents.mark_alert_event_sent(e) end)
```
vicmgs, line 34: Can we make the mark_alert_event_sent happen at the end of the functions send_specific_event_email and send_webhook? that way you break up the total number of events marked sent in a single thread and move them to where the sending event actually happened

ambuznego: yes, good idea!

ambuznego: updated 👆

vicmgs (approved):

vicmgs:
naiiiiisuuuu

PR #666, opened 2021-06-25
Fix subscription so orgs are refetched upon adding new one from top bar

ambuznego:
Previously, the user ID value was being sliced so it did not match the ID that was being used to broadcast a change in orgs. Thus, when user added a new org from top bar, it would not update the list and required a refresh to see the new org show up.


vicmgs (approved):
nice find

PR #646, opened 2021-06-04
Remove unused tables and columns

vicmgs:

PR #676, opened 2021-07-02
Include channel ID in details for notification event

ambuznego:
This fix is needed for the event migration script in https://github.com/helium/console/pull/646 to be successful. Since in 2.0, integrations are decoupled from labels, we need this data to persist in the details so it can be migrated successfully as an integration notification event.


vicmgs (approved):
makes sense

PR #682, opened 2021-07-07
Update toggle CF List UI

vicmgs:


assets/js/components/labels/UpdateLabelModal.jsx
```diff
@@ -220,6 +227,32 @@ class UpdateLabelModal extends Component {
...
           </TabPane>
+          <TabPane tab="CF List" key="cflist">
+            <div style={{ padding: '30px 50px'}}>
+              <div style={{ marginBottom: 20, display: 'flex', flexDirection: 'row', alignItems: 'center' }}>
+                <Switch
+                  onChange={cfListValue => this.setState({ cfListValue })}
+                  checked={cfListValue}
+                  style={{ marginRight: 8 }}
+                />
+                <Text strong style={{ fontSize: 16 }}>Join-Accept CF List Enabled</Text>
```
ambuznego, line 238: On my flows one I was including `(applicable to US915 devices only)` since moogey said to make it clear, should we do that here too?

vicmgs: good point, adding

ambuznego (approved):

Issue #683, opened 2021-07-07
Make Initial organization dc balance as a parameter


disk91:
1. Are you using Helium hosted Console or running your own open source Console?
OSS

2. If you are on Helium hosted Console, which url are you currently on? 
helium-iot.eu

3. If you are experiencing bugs in Helium hosted Console, please specify your Organization name and Device ID (ex. 3483050b-284d-47e2-8eda-e087fc03a753).

Please describe your issue:

Evolution, It would be good to be able to setup the initial org dc_balance with a env parameter
The 10.000 are hardcoded in lib/console_web/controllers/organization_controller.ex and could be replaced by an env variable with a default value 10000.

vicmgs:
ok, we will make this possible after console 2.0 goes to prod

vicmgs:
closing, will be deployed to prod soon

use INITIAL_ORG_GIFTED_DC .env variable

vicmgs:
deployed to prod along with console 2.0, check the latest release notes for upgrade instructions

PR #692, opened 2021-07-14
Set INITIAL_ORG_GIFTED_DC env var

ambuznego:
https://github.com/helium/console/issues/683


templates/.env
```diff
@@ -4,6 +4,7 @@ GUARDIAN_SECRET_KEY=[generate by running: "openssl rand -base64 64"]
 ROUTER_SECRETS=[Unix Epoch Timestamp: openssl rand -base64 64]
 OUI=3
 SELF_HOSTED=true
+INITIAL_DC_BALANCE=10000
```
vicmgs, line 7: what do you think about naming it INITIAL_ORG_GIFTED_DC?

vicmgs (approved):

templates/.env
ambuznego: done!

PR #693, opened 2021-07-14
Chunk massive app.js into separate bundles

vicmgs:
Benefits of chunking like this:
1. Users can download the previously single large js file in chunks in parallel
2. Users can cache these chunks that won't change very often

This chunking setup does not lazy load, so first time users still need to download 8.4MB worth of js, but subsequently visits will be faster due to caching.


ambuznego (approved):
🤞

PR #697, opened 2021-07-15
2.0 fixes for alert + multibuy

ambuznego:
- Multi buy show page should match rest of UI (name input was previously populated vs rest of UI uses the existing name as placeholder)
- Fix skeleton element to show in format of columns for better UX (both alerts and multibuy)
image
- Fix error to show on the form side for better UX (both alerts and multibuy)
image
- Fix typo that caused title for new alert for device/label to not show up


vicmgs (approved):
nice finds!

PR #700, opened 2021-07-16
Fix search for specific entities + modal for adding devices to label

ambuznego:
- Remove `LIMIT 5` from search query for specific entities
- Fix modal CSS to see whole list when it overflows
- Rest of changes are linting


vicmgs (approved):
👍

Issue #707, opened 2021-07-20
Device 'Last Modified' date shows current date/time.


jthiller:
**1. Are you using Helium hosted Console or running your own open source Console?**
Hosted Console

**2. If you are on Helium hosted Console, which url are you currently on?** 
https://staging-console.helium.wtf/flows

**3. If you are experiencing bugs in Helium hosted Console, please specify your Organization name and Device ID (ex. 3483050b-284d-47e2-8eda-e087fc03a753).**
`Joey`
`e9704026-b272-4faa-9e4d-a7ca6a067ac5`

**Please describe your issue:**
'Last Modified' date shows current time without any modifications.

Steps to repro:
1. Visit https://staging-console.helium.wtf/flows
2. Without making any changes, click on an existing device card.
3. Note that the 'Last Modified' time matches current time.

<kbd><img width="427" alt="image" src="https://user-images.githubusercontent.com/1965053/126341860-8c9480f8-8c2f-4731-87df-6bea2d33e96d.png"></kbd>

jthiller:
Please add `Staging` label.

ambuznego:
Thanks for reporting, @jthiller! I'll push a fix soon :)

vicmgs:
fix on staging

PR #709, opened 2021-07-20
Remove special logic for devices w/ no labels + Fixes for last modified on node sidebar

ambuznego:
Fixes issue where "Last Modified" for devices and labels would show current time since the `updated_at` time was not being pulled (https://github.com/helium/console/issues/707)

Removes special logic that dulled the device name and showed None in red when it had no labels (per product request)
image

### DIFF
--- a/assets/js/components/devices/DeviceShow.jsx
+++ b/assets/js/components/devices/DeviceShow.jsx
@@ -254,13 +254,13 @@ class DeviceShow extends Component {
       );
     if (error)
       return (
-        <FunctionDashboardLayout {...this.props}>
+        <DeviceDashboardLayout {...this.props}>
           <div style={{ padding: 40 }}>
             <Text>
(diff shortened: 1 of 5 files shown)


assets/js/components/devices/DeviceShow.jsx
```diff
@@ -254,13 +254,13 @@ class DeviceShow extends Component {
       );
     if (error)
       return (
-        <FunctionDashboardLayout {...this.props}>
+        <DeviceDashboardLayout {...this.props}>
```
vicmgs, line 257: omg wtf.... good catch lol

vicmgs (approved):

assets/js/components/devices/DeviceShow.jsx
ambuznego: no idea how this didn't fail before lol

PR #713, opened 2021-07-22
Reduce docker compose build time

ambuznego:
Total build time w/ new `Dockerfile` (`staging` branch): 259.2s
Total build time w/ prior `Dockerfile` (`staging` branch): 432.1s

Based the changes on the `Dockerfile` off of: https://github.com/phoenixframework/phoenix/blob/master/guides/deployment/releases.md

Tweaked it to prevent permission errors post-release

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,31 +55,10 @@ ARG MIX_ENV
 ENV USER="nobody"
 
 WORKDIR "/home/nobody/app"
-# Creates an unprivileged user to be used exclusively to run the Phoenix app
-# RUN \
-#   addgroup \
-#    -g 1000 \
-#    -S "${USER}" \
-#   && adduser \
-#    -s /bin/sh \
-#    -u 1000 \
-#    -G "${USER}" \
-#    -h /home/nobody \
-#    -D "${USER}" \
-#   && su "${USER}"
-
-# Everything from this line onwards will run in the context of the unprivileged user.
 
 COPY --from=build --chown="${USER}" /app/_build/prod/rel/console ./
 RUN chown -R nobody: .
 USER nobody
 
 ENTRYPOINT ["bin/console"]
-
-# Usage:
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -24,22 +25,40 @@ RUN mix deps.get --only $MIX_ENV
...
 
 # build assets
+COPY priv priv
+
+# Note: if your project uses a tool like https://purgecss.com/,
+# which customizes asset compilation based on what it finds in
+# your Elixir templates, you will need to move the asset compilation step
+# down so that `lib` is available.
 COPY assets assets
-RUN rm assets/webpack.config.js
```
vicmgs, line 28: we need to keep these lines because the webpack docker config needs to replace the regular webpack config due to env vars

vicmgs (commented):
wow looks like magic lolz

Dockerfile
ambuznego: brought them back 👍 check last commit

Dockerfile
```diff
@@ -24,22 +25,43 @@ RUN mix deps.get --only $MIX_ENV
...
+FROM alpine:3.12.1 AS app
+RUN apk add --no-cache libstdc++ openssl ncurses-libs
+
+ARG MIX_ENV
+ENV USER="nobody"
+
+WORKDIR "/home/nobody/app"
+
+COPY --from=build --chown="${USER}" /app/_build/prod/rel/console ./
+RUN chown -R nobody: .
```
michaeldjeffrey, line 63: is this still needed if you're copying as `nobody`?

ambuznego: it builds ok without it but then I get tons of permission errors when trying to run console. was a pain to figure out but this doesn't take nearly as long as before.

vicmgs:
i tried using this when upgrading packages but couldn't get it to work. Closing for now

PR #716, opened 2021-07-22
backend for in xor filter work

ambuznego:
`in_xor_filter` column to be used to display when device has NOT YET entered XOR filter

### DIFF
--- a/lib/console/devices/devices.ex
+++ b/lib/console/devices/devices.ex
@@ -172,4 +172,10 @@ defmodule Console.Devices do
     devices = from(d in Device, where: d.organization_id == ^organization.id) |> Repo.all()
     length(devices)
   end
+
+  def update_in_xor_filter(device_ids) do
+    from(d in Device, where: d.id in ^device_ids) |> Repo.update_all(set: [in_xor_filter: true])
+    devices = from(d in Device, where: d.id in ^device_ids) |> Repo.all()
+    {:ok, devices}
+  end
 end
--- a/lib/console/devices/devices.ex
+++ b/lib/console/devices/devices.ex
@@ -173,9 +173,8 @@ defmodule Console.Devices do
     length(devices)
   end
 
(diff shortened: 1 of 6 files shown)


lib/console_web/controllers/router/device_controller.ex
```diff
@@ -425,4 +425,24 @@ defmodule ConsoleWeb.Router.DeviceController do
...
   end
+
+  def update_devices_in_xor_filter(conn, %{"added" => added_device_ids, "removed" => removed_device_ids}) do
+    result = Ecto.Multi.new()
+      |> Ecto.Multi.run(:added_devices, fn _repo, _ ->
+        with {:ok, devices} <- Devices.update_in_xor_filter(added_device_ids, true) do
+          {:ok, devices}
+        end
+      end)
+      |> Ecto.Multi.run(:removed_devices, fn _repo, _ ->
```
vicmgs, line 436: I think we dont need to remove any devices. Michael is just sending them to cover all bases

lib/console/devices/devices.ex
```diff
@@ -172,4 +172,10 @@ defmodule Console.Devices do
     devices = from(d in Device, where: d.organization_id == ^organization.id) |> Repo.all()
     length(devices)
   end
+
+  def update_in_xor_filter(device_ids, value) do
+    from(d in Device, where: d.id in ^device_ids) |> Repo.update_all(set: [in_xor_filter: value])
```
vicmgs, line 177: Given how we use this function in the controller file, it doesnt seem necessary to do the next line. We can just return the result of Repo.update_all

ambuznego: you is right

lib/console_web/controllers/router/device_controller.ex
ambuznego: oh gotcha. will remove

vicmgs (approved):

PR #720, opened 2021-07-26
Show help links on empty flows page

vicmgs:


assets/js/components/flows/FlowsWorkspace.jsx
```diff
@@ -236,6 +240,28 @@ export default ({
...
+              }}
+            >
+              <div className="pod" id="left">
+                <RocketFilled className="bigicon" />
+                <h2>Get Started with Console</h2>
+                <Link to="/devices"><p>Add and Manage Devices <span>for the Helium Network</span><CaretRightOutlined className="caret" /></p></Link>
+                <Link to="/integrations"><p>Set up an Integration <span>to send and receive device data</span><CaretRightOutlined className="caret" /></p></Link>
+                <Link to="/functions"><p>Apply Functions <span>to your devices</span><CaretRightOutlined className="caret" /></p></Link>
+                <Link to="/users"><p>Invite other Users <span> to your Console Organization</span><CaretRightOutlined className="caret" /></p></Link>
+              </div>
```
ambuznego, line 261: could we extract this from `Welcome.jsx` as well from here into its own component and reuse? would be easier if we ever have to update links or anything here

ambuznego (approved):

PR #738, opened 2021-08-02
Do not run update_xor_filter if length is 0

ambuznego:


vicmgs (approved):
instead of returning {:error, :not_found, "No devices were updated"}

should probably return like a {:error, "No devices were updated"} since its not 404 issue

PR #742, opened 2021-08-03
Send custom xor filter removal error to appsignal

vicmgs:
Creates a way to send "incidents" to appsignal for tracking certain events that shouldn't error out in the conventional way, but we should keep a log of such incidents. Starting off with the removal of existing devices from xor filter for now. We will add more to the "incident" log going forward.

### DIFF
--- a/lib/console_web/controllers/router/device_controller.ex
+++ b/lib/console_web/controllers/router/device_controller.ex
@@ -426,7 +426,7 @@ defmodule ConsoleWeb.Router.DeviceController do
     end
   end
 
-  def update_devices_in_xor_filter(conn, %{"added" => added_device_ids, "removed" => _removed_device_ids}) do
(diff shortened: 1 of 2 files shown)


lib/console_web/controllers/router/device_controller.ex
```diff
@@ -443,5 +443,11 @@ defmodule ConsoleWeb.Router.DeviceController do
     else
       conn |> send_resp(200, "")
     end
+
+    removed_devices = Devices.get_devices_in_list(removed_device_ids)
```
ambuznego, line 447: would we want to do this after the condition so we don't do a db get if unnecessary?

vicmgs: we have to do this to check if the removed_device_ids router sends over exist, the check in the next line is to make sure that this query returns 0 items

ambuznego (approved):

PR #746, opened 2021-08-05
Update time range for clearing sent alert events

vicmgs:


lib/console/jobs.ex
```diff
@@ -91,7 +91,7 @@ defmodule Console.Jobs do
 
   def delete_sent_alerts do
     # since events are kept as "sent" so we can check against flapping, delete them in 24-hr batches
```
ambuznego, line 93: can you update the comment too so we don't think the 25 is a typo lol

vicmgs: good point!

ambuznego (approved):

PR #749, opened 2021-08-06
Email format + minor fixes

ambuznego:
Few things going on here:
- Updating the interval for checking for devices that stop transmitting, should have always been 15 (per comment)
- Updating format in templates to make sure font is same all across
- Update alert last triggered time only when an alert event is actually created (it was previously doing it even if the flapping prevention stopped it from creating one)
- Removing the white space in email preview line
- Removed unnecessary styling so that it inherits instead

tests
```
.......................................................

Finished in 1.9 seconds
55 tests, 0 failures
```

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -64,7 +64,7 @@ config :console,
 config :console, Console.Scheduler,
   jobs: [
     trigger_device_stops_transmitting: [
-      schedule: "*/3 * * * *", # every 15th min
+      schedule: "*/15 * * * *", # every 15th min
       task: {Console.Jobs, :trigger_device_stops_transmitting, []}
     ],
     send_alerts: [
--- a/lib/console/alerts/alert_events.ex
+++ b/lib/console/alerts/alert_events.ex
@@ -57,11 +57,14 @@ defmodule Console.AlertEvents do
           if not email_restrict_to_prevent_flapping do
             email_attrs = Map.put(attrs, :type, "email")
             if Map.has_key?(alert_event_email_config, "value") do
+              # since email and webhook can have diff value, must check against buffer value checked
               if alert_event_email_config["value"] == details.buffer do
                 create_alert_event(email_attrs)
+                Alerts.update_alert_last_triggered_at(alert)
               end
             else
               create_alert_event(email_attrs)
+              Alerts.update_alert_last_triggered_at(alert)
             end
           end
         end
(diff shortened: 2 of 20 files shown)


config/config.exs
```diff
@@ -64,7 +64,7 @@ config :console,
 config :console, Console.Scheduler,
   jobs: [
     trigger_device_stops_transmitting: [
-      schedule: "*/3 * * * *", # every 15th min
```
ambuznego, line 67: this 3 must have snuck in from testing 🤦‍♀️

lib/console/alerts/alert_events.ex
```diff
@@ -78,16 +81,17 @@ defmodule Console.AlertEvents do
...
             webhook_attrs = Map.put(attrs, :type, "webhook")
             if Map.has_key?(alert_event_webhook_config, "value") do
+              # since email and webhook can have diff value, must check against buffer value checked
               if alert_event_webhook_config["value"] == details.buffer do
                 create_alert_event(webhook_attrs)
+                Alerts.update_alert_last_triggered_at(alert)
               end
             else
               create_alert_event(webhook_attrs)
+              Alerts.update_alert_last_triggered_at(alert)
```
ambuznego, line 91: this technically could update twice (once for email and once for webhook) but better than to update when it has not created a new alert event bc of flapping prevention

lib/console/jobs.ex
```diff
@@ -48,7 +48,7 @@ defmodule Console.Jobs do
       end
       recipients = Organizations.get_memberships_by_organization_and_role(alert.organization_id, roles) |> Enum.map(fn (member) -> member.email end)
       details = Enum.map(events, fn (e) -> e.details end)
-      has_hotspot_info = case Enum.find(details, fn d -> d["hotspot"] !== nil end) do
```
ambuznego, line 51: when `hotspot: {}` in event body, it was still considering this true and then had all empty values in the email

lib/console_web/templates/email/api_key_email.html.eex
```diff
@@ -200,9 +200,9 @@
 
         <!-- Create white space after the desired preview text so email clients don’t pull other distracting text into the inbox preview. Extend as necessary. -->
         <!-- Preview Text Spacing Hack : BEGIN -->
-        <div style="display: none; font-size: 1px; line-height: 1px; max-height: 0px; max-width: 0px; opacity: 0; overflow: hidden; mso-hide: all; font-family: sans-serif;">
```
ambuznego, line 203: this commenting is to fix the empty spaces in the subject for all files (this originally came from template)

vicmgs: lmao wtf

vicmgs (approved):
NICEE!

PR #752, opened 2021-08-07
Function current info should show in input when updating

ambuznego:
image

Also updated to be a functional component


vicmgs (approved):
🥳

PR #757, opened 2021-08-10
Fix Safari bug re: search bar not working

ambuznego:
Because of the `noselect` class on the search bar's input component, the search bar is unusable in safari. Turns out this class is not needed to prevent any odd behavior as it is not draggable onto the workspace without it.

Removal of one line is only code change, rest is linting.


vicmgs (approved):
👍 👍

PR #764, opened 2021-08-12
Sync hotspots table w/ API results

ambuznego:
run `mix deps.get` to update packages

this PR adds a job to get hotspot data from blockchain API every hour and upsert results on our `hotspots` table.
dev + stg should be set up using `https://api.helium.wtf/v1` as `BLOCKCHAIN_API_URL`

tests:
```
.......................................................

Finished in 2.2 seconds
55 tests, 0 failures
```


lib/console/blockchain_api.ex
```diff
@@ -0,0 +1,57 @@
...
+           include_404s: true,
+           retry_unknown_errors: true
+         ) do
+      {:ok, %Response{body: body, status_code: status_code}} ->
+        case status_code do
+          200 ->
+            {:ok, body |> Jason.decode!()}
+
+          400 ->
+            IO.puts("Cursor invalid")
```
vicmgs, line 36: what do we want to do with all these logs?

lib/console/jobs.ex
```diff
@@ -158,4 +160,54 @@ defmodule Console.Jobs do
       end
     end)
   end
+
+  def sync_hotspots() do
+    case BlockchainApi.each_page("/hotspots", &process_hotspots/1) do
```
vicmgs, line 165: how many comes back from each page?

lib/console/blockchain_api.ex
ambuznego: oops meant to remove these from wallet work copy pasta

lib/console/blockchain_api.ex
```diff
@@ -0,0 +1,57 @@
+defmodule Console.BlockchainApi do
+  use HTTPoison.Base
+  require HTTPoison.Retry
+  alias HTTPoison.Response
+  alias HTTPoison.Retry
+
+  def each_page(url, fun, cursor \\ nil) do
```
vicmgs, line 7: so we update the db 1 page at a time right? but this is still running on a single thread? do you think it's worthwhile to start a new task per page?

priv/repo/migrations/20210812234934_hotspot_nullable_fields.exs
```diff
@@ -0,0 +1,39 @@
...
+    
+    alter table("hotspots") do
+      modify :height, :integer, null: true
+      modify :location, :string, null: true
+      modify :lat, :decimal, null: true
+      modify :lng, :decimal, null: true
+      modify :short_state, :string, null: true
+      modify :short_country, :string, null: true
+      modify :long_city, :string, null: true
+    end
```
vicmgs, line 25: we need to add a uniqueness check for hotspot_address, my fault for forgetting in my previous PR

vicmgs: "hotspots_address_index" needs to be an UNIQUE index in the db

vicmgs (approved):
what speeeeed

PR #768, opened 2021-08-16
Make self-hosted Console utilize SOCKET_CHECK_ORIGIN

mfalkvidd:
SOCKET_CHECK_ORIGIN is already used elsewhere.

Without this change, websocket connections don't work. They used to
work, but stopped working some time around the v2 upgrade.


vicmgs:
Hi, you are trying to update docker-compose-local.yaml in this PR but that file is meant for running docker locally. 

For server deployments, users should be using docker-compose-server.yaml as the template and the readme tells users to update the socket_check_origin arg in that file after copying.

PR #773, opened 2021-08-16
Hotspot monitoring my tab claimed hotspots

vicmgs:


assets/js/components/coverage/CoverageFollowedTab.jsx
```diff
@@ -1,9 +1,93 @@
...
+import { updateOrganizationHotspot } from '../../actions/coverage'
+import startCase from 'lodash/startCase'
+import { minWidth } from "../../util/constants";
 import HeaderFlag from "../../../img/coverage/followed-tab-header-flag.svg";
+import SelectedFlag from "../../../img/coverage/selected-flag.svg";
+import UnselectedFlag from "../../../img/coverage/unselected-flag.svg";
+import { RedStatusSvg, GreenStatusSvg, OrangeStatusSvg } from './CoverageMainTab'
 
 export default (props) => {
+  const columns = [
```
ambuznego, line 15: was gonna do the extraction as part of my pr already but basically since these (I think) are the same as on the other table, should we start a `constants` file under `coverage/` so we can just have one place to change them all if need be

vicmgs: oh yea, i was gonna do it and forgot lol. I'll make the change

ambuznego (approved):
lgtm! just one comment

PR #777, opened 2021-08-19
Hotspot search

ambuznego:


assets/js/Router.jsx
```diff
@@ -336,7 +336,7 @@ const Router = (props) => {
                                 <Profile user={user} {...props} />
                               )}
                             />
-                            {false && (
+                            {true && (
```
vicmgs, line 339: swap back lool

ambuznego: i cry

vicmgs (approved):
niiice

assets/js/Router.jsx
vicmgs: i literally deployed true to staging too sigh

PR #809, opened 2021-08-30
Update charts package + update charts settings

vicmgs:


assets/js/components/coverage/CoverageHotspotShowStatsTab.jsx
```diff
@@ -128,7 +125,7 @@ export default (props) => {
         if (packetsMap[index]) {
           return Object.values(packetsMap[index]).reduce((a,b) => a + b)
         }
-        return 0
+        return 0.5
```
vicmgs, line 128: 0.5 would count as 0 packets in this case to show the grey bars

ambuznego (approved):

PR #810, opened 2021-08-30
Hotspot marker styling on map

ambuznego:
image

### DIFF
--- a/assets/js/components/common/Mapbox.jsx
+++ b/assets/js/components/common/Mapbox.jsx
@@ -154,7 +154,7 @@ export default (props) => {
             750,
             40,
           ],
-          "circle-opacity": 0.7,
+          "circle-opacity": 0.8,
         },
       });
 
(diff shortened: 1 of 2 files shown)


assets/js/components/common/Mapbox.jsx
```diff
@@ -27,8 +32,13 @@ export default (props) => {
         properties: {
           id: h.hotspot_address,
           name: `${h.hotspot_name}`,
-          description: `description for hotspot ${h.hotspot_address}`,
+          alias: hotspot_alias,
           packetCount: h.packet_count,
+          isFollowed: String(
```
vicmgs, line 37: maybe you can use orgHotspotsMap to check if a hotspot is followed? instead of a linear time filter function

vicmgs: actually doesnt matter since we will paginate followed hotspots list later, nvm

vicmgs (approved):

assets/js/components/common/Mapbox.jsx
ambuznego: oh we are? when we do, will have to change how i check if followed or not then

PR #815, opened 2021-09-01
Sorting coverage tables main and followed

vicmgs:


assets/js/components/coverage/Constants.js
```diff
@@ -107,7 +107,7 @@ export const getColumns = (
               selectHotspotAddress(record.hotspot_address);
             }}
           >
-            {hotspot_alias || startCase(data)}
+            {hotspot_alias ? `${hotspot_alias} (${startCase(data)})` : startCase(data)}
```
vicmgs, line 110: if we want to sort by alias, we need a new column on the table, otherwise it gets too complicated to reason about

lib/console/hotspot_stats/hotspot_stats.ex
```diff
@@ -13,4 +13,168 @@ defmodule Console.HotspotStats do
...
+          COUNT(DISTINCT(device_id)) AS device_count
+        FROM hotspot_stats
+        WHERE organization_id = $1 and reported_at_epoch > $2
+        GROUP BY hotspot_address
+      ) stats
+      LEFT JOIN hotspots h ON stats.hotspot_address = h.address
+      LEFT JOIN organization_hotspots os ON stats.hotspot_address = os.hotspot_address
+      ORDER BY
+        CASE $4 WHEN 'asc' THEN
+          CASE $3
```
vicmgs, line 43: has to split up the queries because you cant mix integer columns and string columns for sorting in this case statement

ambuznego (approved):

PR #817, opened 2021-09-02
Map location issues

ambuznego:
This PR:
- Handles NaN coordinates for hotspot markers
- Show location on table when only city + country are provided


vicmgs (approved):
CLEANNNN

PR #820, opened 2021-09-03
Add pagination to hotspotShowDevicesHeardTab

vicmgs:


assets/js/components/coverage/CoverageHotspotShowDevicesTab.jsx
```diff
@@ -34,14 +34,25 @@ const columns = [
 ];
 
 export default (props) => {
+  const [page, setPage] = useState(1);
+  const [pageSize, setPageSize] = useState(25);
```
ambuznego, line 38: this probs shouldn't be in state but instead a const?

ambuznego (approved):
just one thing, otherwise LGTM

assets/js/components/coverage/CoverageHotspotShowDevicesTab.jsx
vicmgs: fixed

PR #825, opened 2021-09-07
Paginate main and followed coverage tabs

vicmgs:


lib/console/hotspot_stats/hotspot_stats_resolver.ex
```diff
@@ -245,7 +254,8 @@ defmodule Console.HotspotStats.HotspotStatsResolver do
           short_state: Enum.at(r, 7),
           latitude: Enum.at(r, 8),
           longitude: Enum.at(r, 9),
-          alias: Enum.at(r,10)
+          alias: Enum.at(r,10),
+          total_entries: Enum.at(r, 11)
```
ambuznego, line 258: can do this later but maybe we can refactor so the main object has the no. of entries and not every entry object

ambuznego (approved):

PR #829, opened 2021-09-08
Mapbox loading issue

ambuznego:
Adding the empty data resource at first and setting the data for the points on the map were _sometimes_ hitting a race condition, resulting in the points sometimes not rendering on the first load. This PR addresses the issue by delaying setting the data by 2 seconds in the edge case that the resource is not finished being added to the map. In my opinion, the extra 2 second does not affect the UX since it happens while the map is resetting to recenter based on the data points.


vicmgs (approved):
lgtm, we can tweak further in the future if needed

PR #863, opened 2021-09-23
Fix bug with coverage organization hotspot aliases

vicmgs:


priv/repo/migrations/20210923231044_remove_status_non_null_from_hotspots.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Console.Repo.Migrations.RemoveStatusNonNullFromHotspots do
+  use Ecto.Migration
+
+  def change do
+    alter table(:hotspots) do
+      modify(:status, :string, null: true)
```
vicmgs, line 6: noticed that we can get status of null from blockchain api, and upon encountering said row, will stop the job completely

lib/console/hotspot_stats/hotspot_stats.ex
```diff
@@ -39,7 +39,7 @@ defmodule Console.HotspotStats do
         GROUP BY hotspot_address
       ) stats
       LEFT JOIN hotspots h ON stats.hotspot_address = h.address
-      LEFT JOIN organization_hotspots os ON stats.hotspot_address = os.hotspot_address
+      LEFT JOIN organization_hotspots os ON stats.hotspot_address = os.hotspot_address and os.organization_id = $1
```
vicmgs, line 42: aliases of the same hotspot from other orgs is getting into the current org's table

PR #867, opened 2021-09-27
Select to claim/unclaim hotspots

ambuznego:
image
image
image

### DIFF
--- a/assets/js/components/coverage/CoverageFollowedTab.jsx
+++ b/assets/js/components/coverage/CoverageFollowedTab.jsx
@@ -1,46 +1,55 @@
 import React, { useEffect, useState } from "react";
 import { Typography, Row, Col, Table, Pagination } from "antd";
 const { Text } = Typography;
-import { updateOrganizationHotspot } from '../../actions/coverage'
-import { getColumns } from './Constants'
-import { minWidth } from "../../util/constants";
+import {
+  updateOrganizationHotspot,
+  updateOrganizationHotspots,
+} from "../../actions/coverage";
+import { ClaimButton, UnclaimButton, getColumns } from "./Constants";
 import HeaderFlag from "../../../img/coverage/followed-tab-header-flag.svg";
+import UserCan from "../common/UserCan";
 
 export default (props) => {
   const [page, setPage] = useState(1);
   const pageSize = 25;
   const [column, setColumn] = useState("packet_count");
   const [order, setOrder] = useState("desc");
+  const [selectedRows, setSelectedRows] = useState([]);
+  const [allSelected, setAllSelected] = useState(false);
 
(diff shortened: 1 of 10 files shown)


assets/js/components/coverage/CoverageFollowedTab.jsx
```diff
@@ -1,9 +1,8 @@
 import React, { useEffect, useState } from "react";
 import { Typography, Row, Col, Table, Pagination } from "antd";
 const { Text } = Typography;
-import { updateOrganizationHotspot } from '../../actions/coverage'
-import { getColumns } from './Constants'
-import { minWidth } from "../../util/constants";
+import { updateOrganizationHotspot } from "../../actions/coverage";
```
ambuznego, line 4: only linting done to this file

lib/console/organization_hotspots/organization_hotspots.ex
```diff
@@ -34,4 +34,28 @@ defmodule Console.OrganizationHotspots do
...
+          Repo.insert!(OrganizationHotspot.changeset(%OrganizationHotspot{}, attrs))
+        end)
+      end)
+    do
+      {:ok, length(new_organization_hotspots), new_organization_hotspots}
+    end
+  end
+
+  def unclaim_org_hotspots(hotspot_addresses, organization) do
+    query = from(oh in OrganizationHotspot, where: oh.hotspot_address in ^hotspot_addresses and oh.organization_id == ^organization.id)
```
vicmgs, line 58: we dont want to simply delete them all since we could have a row that isn't claimed but has the alias set

ambuznego: ah! good catch. will update.

lib/console/organization_hotspots/organization_hotspots.ex
```diff
@@ -34,4 +34,28 @@ defmodule Console.OrganizationHotspots do
   def delete_org_hotspot(%OrganizationHotspot{} = org_hotspot) do
     Repo.delete(org_hotspot)
   end
+
+  def claim_org_hotspots(hotspot_addresses, organization) do
+    new_organization_hotspots = Enum.reduce(hotspot_addresses, [], fn hotspot_address, acc ->
+      if Repo.get_by(OrganizationHotspot, hotspot_address: hotspot_address, organization_id: organization.id) == nil do
```
vicmgs, line 40: can we run 1 query to find all the OrgHotspots and reduce through the results to create the list of attrs?

Also, a hotspot may not be nil even if it's not claimed because it could have an alias

vicmgs: just the first part of my comment then!

vicmgs (approved):
nicely done

PR #870, opened 2021-09-27
Add search by wallet capability

ambuznego:
image
(removed wallet id post-screenshot)


priv/repo/migrations/20210927215204_add_owner_to_hotspots_table.exs
```diff
@@ -0,0 +1,17 @@
+defmodule Console.Repo.Migrations.AddOwnerToHotspotsTable do
+  use Ecto.Migration
+
+  def up do
+    alter table(:hotspots) do
+      add :owner, :string, null: true
+      modify :status, :string, null: true
```
vicmgs, line 7: do we need the status modification here?

vicmgs: see previous migration before this one you are trying to commit

PR #924, opened 2021-11-08
Hotspot groups

ambuznego:
Currently, only followed hotspots can be followed. Unfollowing a hotspot that is in a group will also remove it from the group.

image
image
image
image
image
image

image
image


priv/repo/migrations/20211020174207_hotspot_groups.exs
```diff
@@ -0,0 +1,23 @@
...
+
+    create table(:hotspots_groups) do
+      add :group_id, references(:groups)
+      add :hotspot_id, references(:hotspots)
+
+      timestamps()
+    end
+
+    create unique_index(:hotspots_groups, [:group_id, :hotspot_id])
+    create unique_index(:groups, [:name, :organization_id])
```
vicmgs, line 21: do we need any more indices to improve query performance? For example, we might want an index on the groups table for organization_id, or an index on hotspots_groups table for organization_id etc.

ambuznego: org id index added on groups

vicmgs (approved):

PR #933, opened 2021-11-22
Fix response body issue, coverage stats query + search columns

vicmgs:


lib/console_web/controllers/router/device_controller.ex
```diff
@@ -185,6 +185,19 @@ defmodule ConsoleWeb.Router.DeviceController do
         result =
           Ecto.Multi.new()
           |> Ecto.Multi.run(:event, fn _repo, _ ->
+            event = case event["data"]["res"]["body"] do
+              nil -> event
+              _ ->
+                cond do
+                  String.length(event["data"]["res"]["body"]) > 2500 ->
+                    Kernel.put_in(event["data"]["res"]["body"], "Request body is too long and cannot be displayed properly.")
```
ambuznego, line 193: should say `Response`

ambuznego (approved):
just one minor fix, otherwise LGTM

lib/console_web/controllers/router/device_controller.ex
vicmgs: so derp.... thanks!

PR #943, opened 2021-12-01
Update v1 controller responses for adr and cf list

vicmgs:


lib/console/devices/devices.ex
```diff
@@ -35,6 +35,14 @@ defmodule Console.Devices do
     query |> Repo.all()
   end
 
+  def list_devices_no_disco_mode() do
```
vicmgs, line 38: adding this back because one of the migrations needs this function

ambuznego (approved):

PR #951, opened 2021-12-06
Show inherited config profiles on device show details

vicmgs:


assets/js/components/devices/DeviceShow.jsx
```diff
@@ -577,7 +578,24 @@ export default (props) => {
                               </Link>
                             ) : (
                               <Text>
-                                <i>None selected</i>
+                                <i>None </i>
```
ambuznego, line 581: shouldn't this be showing the profile name and not none if there is one

assets/js/components/flows/infoSidebar/DeviceContent.jsx
```diff
@@ -406,6 +407,31 @@ class DeviceContent extends Component {
                   OTAA
                 </Tag>
               </Paragraph>
+              <Paragraph>
+                <Text strong>Profile: </Text>
+                { device.config_profile_id ? (
+                  <Text>{device.config_profile.name}</Text>
+                ) : (
+                  <Text>
+                    None
```
ambuznego, line 416: same here

PR #972, opened 2021-12-17
Toggle intercom button selectively in mobile mode

vicmgs:


assets/js/actions/display.js
```diff
@@ -1,6 +1,14 @@
 export const UPDATE_DISPLAY_SETTING = 'UPDATE_DISPLAY_SETTING';
 
 export const updateDisplay = (desktopOnly) => {
+  if (desktopOnly) {
+    window.Intercom('update', { "hide_default_launcher": false });
+  } else {
+    if (window.innerWidth < 721) {
+      window.Intercom('update', { "hide_default_launcher": true });
+    }
```
ambuznego, line 9: would there ever be a non-desktop instance where innerWidth > 721 that we would still want to hide?

vicmgs: not for now, anything over 721 will always go to desktop mode

ambuznego: do we still need the if then?

vicmgs: its an edge case of:

Your browser window is small.
You switch to desktop only.
You make the screen size > 721.
You click the mobile mode button but your screen size still > 721.

ambuznego (approved):
sendit

vicmgs:
much thanks for monstrous catch

PR #976, opened 2021-12-21
Add mobile tag for analytics

ambuznego:


assets/js/components/Welcome.jsx
```diff
@@ -1,173 +1,352 @@
...
-    hideWelcomeScreen: localStorage.getItem('hideWelcomeScreen')
-  }
+    hideWelcomeScreen: localStorage.getItem("hideWelcomeScreen"),
+  };
 
   componentDidMount() {
-    analyticsLogger.logEvent("ACTION_NAV_WELCOME_PAGE")
+    analyticsLogger.logEvent(
+      isMobile ? "ACTION_NAV_WELCOME_PAGE_MOBILE" : "ACTION_NAV_WELCOME_PAGE"
+    );
```
ambuznego, line 24: main change here, rest is linting

assets/js/components/devices/DeviceNew.jsx
```diff
@@ -112,13 +113,16 @@ class DeviceNew extends Component {
...
     if (devEUI.length === 16 && appEUI.length === 16 && appKey.length === 32) {
-      analyticsLogger.logEvent("ACTION_CREATE_DEVICE", {
-        name: name,
-        devEUI: devEUI,
-        appEUI: appEUI,
-        appKey: appKey,
-        configProfileId: configProfileId,
-      });
+      analyticsLogger.logEvent(
+        isMobile ? "ACTION_CREATE_DEVICE_MOBILE" : "ACTION_CREATE_DEVICE",
```
ambuznego, line 117: main change here, rest is linting

assets/js/util/constants.js
```diff
@@ -2,3 +2,5 @@ export const minWidth = 700;
 
 export const checkIfDevicesNotInFilter = (label) =>
   label.devices.filter((device) => device.in_xor_filter === false).length > 0;
+
+export const isMobile = window.innerWidth < 721 || window.innerHeight < 721;
```
vicmgs, line 6: we may want to drop the innerHeight part now that I think about it since the mobile view wouldn't show just because height is low.

Instead we should check for desktop only from store state?

vicmgs (approved):

PR #990, opened 2022-01-11
Add connection to amqp for events endpoint

vicmgs:


lib/console/etl_worker.ex
```diff
@@ -0,0 +1,305 @@
...
+    GenServer.start_link(__MODULE__, initial_state, name: __MODULE__)
+  end
+
+  def init(_opts) do
+    schedule_events_etl(100)
+    {:ok, %{}}
+  end
+
+  def handle_info(:run_events_etl, state) do
+    Task.Supervisor.async_nolink(ConsoleWeb.TaskSupervisor, fn ->
```
ambuznego, line 19: we're using a task for this so it runs async w/ other jobs like sending emails. multiple `handle_info` won't be running at once, right?

lib/console_web/controllers/router/device_controller.ex
```diff
@@ -163,6 +163,95 @@ defmodule ConsoleWeb.Router.DeviceController do
   end
 
   def add_device_event(conn, %{"device_id" => device_id} = event) do
+    if true do
```
ambuznego, line 166: should we update this to use an env variable so it does not run w/ rabbitmq for undesired environments or open source?

lib/console/etl_worker.ex
vicmgs: only 1 handle info will run at a time.

running a task here so that if it fails, it doesnt take down the entire etl_worker genserver.

vicmgs:
Env vars for amqp commit changes dev to NOT use amqp by default. Change use_amqp_events to "true" in dev.exs locally to test

PR #1003, opened 2022-01-27
Change flow when user with no org logs in but has pending invitation(s)

ambuznego:


lib/console/email.ex
```diff
@@ -86,6 +86,19 @@ defmodule Console.Email do
     |> render(:invitation_email)
   end
 
+  def resend_invitation_email(%Invitation{email: email, token: token, role: role}, inviter_email, %Organization{name: organization_name}) do
```
ambuznego, line 89: same function as above but since the invitee is the one logged in, I don't have `conn.current_user` to pass as `User%{}` so using the email directly here.

assets/js/actions/invitation.js
```diff
@@ -1,28 +1,37 @@
...
+  };
+};
+
+export const getInvitations = async (email) => {
+  const invitations = await rest.get(`/invitations/${email}`);
+  return invitations.data;
+};
+
+export const resendInvitations = async (email) => {
+  const result = await rest.get(`/resend_invitation/${email}`);
```
vicmgs, line 35: maybe this should be a post? Get seems like the wrong request type here

assets/js/components/organizations/NoOrganization.jsx
```diff
@@ -1,78 +1,180 @@
...
 
-  handleInputUpdate = (e) => {
-    this.setState({ [e.target.name]: e.target.value})
-  }
+  const fetchInvitations = async () => {
+    getInvitations(user.email)
+      .then((invitations) => {
+        setInvitations(invitations);
+      })
+      .finally(() => setInvitationsLoading(false));
```
vicmgs, line 24: Oooo something new to me

lib/console/email.ex
vicmgs: You can just create a fake %User{email: inviter_email} object where this is being called and you can use the above function

lib/console/organizations/organizations.ex
```diff
@@ -363,4 +363,23 @@ defmodule Console.Organizations do
...
+
+  def get_inviter_email(user_id) do
+    memberships = from(m in Membership, where: m.user_id == ^user_id)
+    |> Repo.all()
+
+    List.first(memberships).email
+  end
+
+  def get_latest_invitation(email) do
+    invitations = from(i in Invitation, where: i.email == ^email, order_by: [desc: :inserted_at])
```
vicmgs, line 380: minor optimization, add limit 1 to query

assets/js/actions/invitation.js
ambuznego: ah shoot i was getting an error w/ and switched to get just to get it working and then forgot. totally agree.

vicmgs (approved):
small comments, much great!

PR #1024, opened 2022-02-03
Fix tests

ambuznego:
```
➜  console git:(fixTests) mix test

14:35:28.095 [info]  Migrations already up
............................................................

Finished in 2.1 seconds (0.5s async, 1.6s sync)
60 tests, 0 failures

Randomized with seed 842007
```


lib/console_web/controllers/invitation_controller.ex
```diff
@@ -14,21 +14,28 @@ defmodule ConsoleWeb.InvitationController do
   def create(conn, %{"invitation" => attrs}) do
     current_user = conn.assigns.current_user
     current_organization = conn.assigns.current_organization
-    organization = Organizations.get_organization!(current_user, attrs["organization"])
 
-    if current_organization.id == organization.id and current_user.email != attrs["email"] do
-      with {:ok, %{ invitation: invitation, user: _ }} <-
```
ambuznego, line 20: using this `with` clause, we weren't catching the error when the invitation (or user) fails to be created and thus the whole thing is rolled back

vicmgs: what do you mean? would Organizations.create_invitation return {:ok , _} if it fails?

ambuznego: we'd get `no function clause matching` error

vicmgs (approved):

PR #1030, opened 2022-02-08
Add controller actions to v1 for integrations and flows

vicmgs:


lib/console_web/controllers/v1/flow_controller.ex
```diff
@@ -0,0 +1,122 @@
...
+                device_ids =
+                  Devices.get_devices_for_label(label.id)
+                  |> Enum.map(fn d -> d.id end)
+
+                broadcast_router_update_devices(device_ids)
+
+                conn
+                |> send_resp(:ok, "Flow with integration created successfully")
+              end
+            end
```
ambuznego, line 88: could this be simpler by conditionally adding `label_id`, `device_id`, and `function_id` to a variable, say we call it `params`, and then just calling `Flows.create_flow(params)` in one same `with` clause? may be easier to read and less same lines of code

lib/console_web/views/v1/label_view.ex
```diff
@@ -14,10 +14,13 @@ defmodule ConsoleWeb.V1.LabelView do
     %{
       id: label.id,
       name: label.name,
-      adr_allowed: label.adr_allowed,
-      cf_list_enabled: label.cf_list_enabled,
-      rx_delay: label.rx_delay,
-      config_profile_id: label.config_profile_id
```
ambuznego, line 20: why are we removing these?

vicmgs: oh crap, nice catch

lib/console_web/controllers/v1/flow_controller.ex
vicmgs: let me try, my brain wasn't super on point earlier

ambuznego (approved):

PR #1040, opened 2022-02-14
Device v1 controller set active by device creds

vicmgs:


lib/console_web/router.ex
```diff
@@ -139,8 +139,10 @@ defmodule ConsoleWeb.Router do
 
     get "/organization", OrganizationController, :show
     resources "/devices", DeviceController, only: [:index, :show, :create, :delete, :update]
+    post "/devices/active", DeviceController, :set_devices_active
```
ambuznego, line 142: should use `put` instead of `post` for both

PR #1062, opened 2022-02-26
Improve Azure credentials form to prevent user error

ambuznego:
image

Upon entering the connection string, the other fields populate:
image


michaeldjeffrey:
Do you want to leave the same little `i` icon for the Hubname? In case someone thinks their connection string was parsed incorrectly and go to update by hand.

ambuznego:
> Do you want to leave the same little `i` icon for the Hubname? In case someone thinks their connection string was parsed incorrectly and go to update by hand.

As discussed, the rest of the fields are disabled so user must enter the connection string, ideally preventing any user error since this info must match Azure.

vicmgs (approved):

PR #1064, opened 2022-03-01
Add function v1 controller actions + create custom integrations

vicmgs:


lib/console_web/controllers/v1/channel_controller.ex
```diff
@@ -134,6 +134,111 @@ defmodule ConsoleWeb.V1.ChannelController do
...
+          "aws_secret_key" => sk,
+          "aws_region" => region
+        },
+        "name" => name,
+        "type" => "aws",
+        "organization_id" => current_organization.id
+      }
+
+    with {:ok, %Channel{} = channel} <- Channels.create_channel(current_organization, channel_params) do
+      channel = attach_devices_and_labels(current_organization, channel)
```
ambuznego, line 154: there wouldn't be any devices or labels attached if I just created the channel, right?

ambuznego (approved):

PR #1073, opened 2022-03-09
Add ability to import / export org

vicmgs:


assets/js/components/organizations/NewOrganizationModal.jsx
```diff
@@ -52,6 +59,63 @@ class NewOrganizationModal extends Component {
           onChange={this.handleInputUpdate}
           style={{ marginBottom: 20 }}
         />
+        <Text>or</Text>
+
+        {
+          this.state.importFailed ? (
```
ambuznego, line 65: we should we be resetting `importFailed` to `false` when closing the modal otherwise when I reopen the modal, it will still show the error msg
image

ambuznego (approved):

PR #1086, opened 2022-03-23
Small fixes for Coverage

ambuznego:
- Previously, when the search was loading, it would show NULL in the pagination dropdown
- Sometimes when refetching results with search, loading wouldn't be true unless the prop is added to show loading on network status change


vicmgs (approved):
love it, good find

Crisatian:
Dose not sync with blockchin, legs bihaind(20, 50 block)

PR #1092, opened 2022-03-28
Increase multibuy to 16

Oliv4945:
This pull request increases the maximum number of packets that can be bought to 16.  
This is useful for RSSI location services like Semtech's [LoRa Cloud](https://www.loracloud.com/) where the accuracy is increased when you send up to 16 packets. Today one needs to buy all available packets to benefit of this but it can quickly become expensive for example in the Bay area where a packet is received by 300 or 400 gateways...

The change
image

Example of location, in yellow the RSSI loc when buying 9 packets
image

### DIFF
--- a/assets/js/components/multi_buy/MultiBuyForm.jsx
+++ b/assets/js/components/multi_buy/MultiBuyForm.jsx
@@ -136,7 +136,7 @@ export default ({ show, id, openDeleteMultiplePacketModal }) => {
                   <Slider
                     value={multiBuyValue}
                     min={1}
-                    max={10}
+                    max={17}
                     tooltipVisible={false}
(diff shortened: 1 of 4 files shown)


jdgemm:
Thanks we'll review with team...

jdgemm:
Thanks for the pr, we're going to make the number 20 and still include an all available option.

vicmgs:
recreating our own PR with 20 as max

Oliv4945:
Thanks @jdgemm @vicmgs  for this, what is the rational behind selecting the value of 20?

jdgemm:
Providing a buffer for users who want up to 20, but not all...

PR #1095, opened 2022-03-29
Add ability to limit devices and orgs 

vicmgs:


lib/console/devices/devices.ex
```diff
@@ -114,23 +114,25 @@ defmodule Console.Devices do
 
   def create_device(attrs \\ %{}, %Organization{} = organization) do
     count = get_organization_device_count(organization)
+
     cond do
-      organization.name !== "Discovery Mode (Helium)" and count >= Application.get_env(:console, :max_devices_in_org) ->
```
ambuznego, line 118: `organization.name !== "Discovery Mode (Helium)"` must stay if we're keeping disco mode code, right?

ambuznego: oh just kidding, adding new function for create discovery device

ambuznego (approved):
I'd suggest disabling or removing Add Organization button. otherwise LGTM!

PR #1110, opened 2022-04-09
Fix CFList Copy. Minor .env updates

lthiery:
Copy did not reflect behavior since Router does no longer alternate CFList anymore. In addition, whether or not Channel Mask is transmitted when CFList is enabled was previously not clear.


lthiery:
Woops I snuck in that Mailgun stuff too. :shrug:

lthiery:
While I was at it, I added some notes abuot Auth0 vs Mailgun. Please review for accuracy @vicmgs

PR #1139, opened 2022-04-27
Help text for device migration

ambuznego:
This PR:
- Allows user to pick whether they want to keep devices active upon exporting their organization (and in turn exporting them as inactive so they are not active on both environments) or deactivate them in the current environment (and in turn exporting them as active so they can be ready on the new environment)
- Shows help text for user to understand the need for device deactivation at both ends of the flow (export + import)
_______________________________________
### Screenshots
image
image
image

### DIFF
--- a/assets/js/components/organizations/OrganizationsTable.jsx
+++ b/assets/js/components/organizations/OrganizationsTable.jsx
@@ -208,7 +208,6 @@ class OrganizationsTable extends Component {
                     icon={<ExportOutlined />}
                     size="small"
                     style={{ marginLeft: 8 }}
-                    // onClick={() => exportOrganization(record.id, record.name)}
                     onClick={() => {
                       this.setState({ showDevicesConfirmModal: true });
                     }}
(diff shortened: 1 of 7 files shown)


assets/js/components/organizations/OrganizationsTable.jsx
```diff
@@ -206,7 +208,10 @@ class OrganizationsTable extends Component {
                     icon={<ExportOutlined />}
                     size="small"
                     style={{ marginLeft: 8 }}
-                    onClick={() => exportOrganization(record.id, record.name)}
+                    // onClick={() => exportOrganization(record.id, record.name)}
```
vicmgs, line 211: leftover line

vicmgs (approved):
amassssing

assets/js/components/organizations/OrganizationsTable.jsx
ambuznego: oopsie

PR #1141, opened 2022-04-29
Add region to downlink

ambuznego:
image

```
➜  console git:(downlinkQueueRegion) ✗ mix test

12:21:11.108 [info]  Migrations already up
......................................................................

Finished in 2.1 seconds (0.6s async, 1.5s sync)
70 tests, 0 failures
```

### DIFF
--- a/assets/js/components/common/Downlink.jsx
+++ b/assets/js/components/common/Downlink.jsx
@@ -10,7 +10,9 @@ import {
   Checkbox,
   Tooltip,
   Button as RegularButton,
+  Select,
 } from "antd";
+const { Option } = Select;
 import DownlinkImage from "../../../img/downlink.svg";
 import ClearOutlined from "@ant-design/icons/ClearOutlined";
 import ReloadOutlined from "@ant-design/icons/ReloadOutlined";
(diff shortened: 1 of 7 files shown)


assets/js/components/common/Downlink.jsx
```diff
@@ -30,6 +33,7 @@ class Downlink extends Component {
     position: "last",
     showRefresh: false,
     queue: [],
+    region: "undefined",
```
vicmgs, line 36: maybe change this string to "default" instead of "undefined"? kinda weird since undefined is an actual entity in js

vicmgs:
does v1 controller not need to change at all?

assets/js/components/common/Downlink.jsx
ambuznego: I had "default" originally but looks like router is expecting undefined: https://discord.com/channels/404106811252408320/966788666913611867/969664709445967952 but I can confirm with Pierre

ambuznego:
> does v1 controller not need to change at all?

Wasn't sure if it was in scope but will check with Pierre

assets/js/components/common/Downlink.jsx
ambuznego: addressed in latest comit

vicmgs (approved):
naice

PR #1160, opened 2022-05-17
Change text

ambuznego:
image
image

### DIFF
--- a/assets/js/components/common/DashboardLayout.jsx
+++ b/assets/js/components/common/DashboardLayout.jsx
@@ -19,7 +19,9 @@ class DashboardLayout extends Component {
     this.setState({ showNav: !this.state.showNav });
   };
   toggleSurveyNotification = () => {
-    this.setState({ showSurveyNotification: !this.state.showSurveyNotification });
+    this.setState({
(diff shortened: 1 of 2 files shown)


vicmgs (approved):

assets/js/components/common/DashboardLayout.jsx
```diff
@@ -171,7 +179,7 @@ class DashboardLayout extends Component {
                       </a>
                     ))}
                     <div style={{ marginLeft: "auto" }}>
-                      © 2021 Helium Systems Inc.
+                      © 2022 Helium Systems Inc.
```
JayKickliter, line 182: If incrementing the date, then should it be Nova?

ambuznego: Will defer to @jdgemm

JayKickliter: Yeah, I don't really know, but figured Helium Systems Inc. wasn't it

ambuznego: per Dal, keeping as is

PR #1183, opened 2022-06-02
Fix double saving when nodes dropped into flows page

vicmgs:
also removes unused hasChanges props passing

### DIFF
--- a/assets/js/components/flows/FlowsWorkspace.jsx
+++ b/assets/js/components/flows/FlowsWorkspace.jsx
@@ -194,7 +194,6 @@ export default ({
     if (type === "functionNode") {
       data = Object.assign({}, data, {
         format: event.dataTransfer.getData("node/function_format"),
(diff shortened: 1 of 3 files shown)


assets/js/components/flows/FlowsWorkspace.jsx
```diff
@@ -187,36 +187,37 @@ export default ({
...
         type: event.dataTransfer.getData("node/channel_type"),
         hasAlerts: event.dataTransfer.getData("node/has_alerts") === "true",
+        lastErrored: event.dataTransfer.getData("node/last_errored") === "true",
       });
     }
 
     if (type === "functionNode") {
       data = Object.assign({}, data, {
         format: event.dataTransfer.getData("node/function_format"),
+        hasAlerts: event.dataTransfer.getData("node/has_alerts") === "true",
```
ambuznego, line 197: We don't have function alerts though

vicmgs: good point, fixed

PR #1186, opened 2022-06-06
Add task await for recurring jobs

vicmgs:


ambuznego (approved):

config/config.exs
```diff
@@ -68,8 +68,9 @@ config :console, Console.Scheduler,
       task: {Console.Jobs, :send_survey_tokens, []}
     ],
     refresh_materialized_views: [
-      schedule: "0 * * * *", # every hour @ 0 mins
-      task: {Console.Jobs, :refresh_materialized_views, []}
+      schedule: "36 * * * *", # every hour @ 36 mins to refresh
```
vicmgs, line 71: Using minute 36 here because we dont want it to happen at the top of the hour since some jobs get triggered at 00.
Also not on every 5 mins since other jobs get triggered every 5 mins.

Basically setting it to 36 to avoid concurrently running other jobs that may screw this up.

PR #1201, opened 2022-06-22
Location-based IP Filtering for Restricted Areas

ambuznego:


config/prod.exs
```diff
@@ -106,8 +106,7 @@ config :console,
 config :console,
   recaptcha_secret_key: System.get_env("RECAPTCHA_SECRET_KEY")
 
-# config :prometheus, App.PrometheusExporter,
-#   path: "/metrics",
-#   format: :auto,
-#   registry: :default,
-#   auth: {:basic, System.get_env("PROMETHEUS_USERNAME"), System.get_env("PROMETHEUS_PASSWORD")}
+config :console,
```
vicmgs, line 109: might not need these here at all since we won't be using them

vicmgs (approved):

PR #1206, opened 2022-06-27
Refactor coverage search

vicmgs:
Refactoring coverage search for better performance.

1. Removed device and packet stats from search results as it makes the query too long and complicated. Also, if a hotspot has transferred data for the org, it will already show up in the main coverage tab.
2. Removed query searches for length < 3, these queries are useless
3. Queries clearly meant for owner wallet search should be optimized


ambuznego (approved):
v nice, one comment

PR #1207, opened 2022-06-27
Gracefully handle when Magic fails to get user metadata

ambuznego:
Currently, when the Magic method for grabbing the user's metadata fails, user sees a loading screen indefinitely. This PR aims to handle this case by logging user out and displaying an error. Currently `console.log`ing the error to see if we can better troubleshoot/debug this issue.

### DIFF
--- a/assets/js/actions/auth.js
+++ b/assets/js/actions/auth.js
@@ -1,72 +1,80 @@
-import * as rest from '../util/rest';
-import { intercomResizeListen } from '../util/intercom'
-import { logout } from '../components/auth/Auth0Provider';
-import { logoutUser } from './magic'
-import analyticsLogger from '../util/analyticsLogger';
-import { config } from '../config/magic'
-import crypto from "crypto"
+import * as rest from "../util/rest";
+import { intercomResizeListen } from "../util/intercom";
+import { logout } from "../components/auth/Auth0Provider";
+import { logoutUser } from "./magic";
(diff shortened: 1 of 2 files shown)


assets/js/actions/auth.js
```diff
@@ -1,72 +1,80 @@
...
 
 export const magicLogIn = (user) => {
-  const hash = crypto.createHmac('sha256', process.env.INTERCOM_ID_SECRET || 'key').update(user.email).digest('hex')
-  window.Intercom('boot', {
-    app_id: 'uj330shp',
-    email: user.email,
-    user_hash: hash
-  })
-  intercomResizeListen()
+  if (user.email) {
```
ambuznego, line 54: main change here is adding this condition and logging out if not met (rest is linting)

assets/js/actions/magic.js
```diff
@@ -1,100 +1,131 @@
...
 
 export const checkUser = async () => {
-  const isLoggedIn = await magic.user.isLoggedIn();
-  if (isLoggedIn) {
-    const user = await magic.user.getMetadata();
-    let needRegistration = false
-    if (window.user_invite_only === 'true' || process.env.USER_INVITE_ONLY === 'true') {
-      needRegistration = await checkRegisteredUser(user.email)
-    }
+  try {
```
ambuznego, line 17: main change is wrapping `checkUser` and `loginUser` in try/catch

PR #1213, opened 2022-06-30
Preferred Hotspot warning when not applicable bug

ambuznego:


assets/js/components/coverage/Constants.js
```diff
@@ -292,8 +293,10 @@ export const ActionButton = ({
         } else if (e.key === "prefer") {
           preferHotspots(selectedAddresses, true);
         } else {
+          console.log(preferredHotspotAddresses);
```
vicmgs, line 296: remvoe log

assets/js/components/coverage/Constants.js
```diff
@@ -92,6 +92,7 @@ export const getColumns = (
                 if (
                   warnUnfollow &&
                   !!hotspot_claimed &&
```
vicmgs, line 94: maybe this should check for hotspot_preferred instead of hotspot_claimed? Since this can only run if the hotspot is preferred right?

vicmgs (approved):
just one comment, may or may not matter

PR #1216, opened 2022-07-05
Remove alert node for manually deleted device when checking for device stops transmitting

ambuznego:
Prevent the following:
```
function nil.active/0 is undefined
```

when `check_device_stops_transmitting` is ran on a deleted device because the alert node was not also deleted.


vicmgs (approved):
❤️

PR #1220, opened 2022-07-08
Enable Flows multi-selection area

ambuznego:
image


vicmgs:
2 comments:

1. Can we move the ? tooltip next to the NODES text?
2. Can multiselecting a single node not toggle the sidepane by listening for the shift key somehow?

PR #1223, opened 2022-07-08
Limit ability to add organization in Nova console

ambuznego:


vicmgs:
we can probably surface the "vetted" attribute through some newly created endpoint that is called on organizationIndex page if USER_INVITE_ONLY?

lib/console/auth/auth.ex
```diff
@@ -16,13 +16,15 @@ defmodule Console.Auth do
   end
 
   def get_user_by_id_and_email(user_id, email) do
+    user = get_user_by_email(email)
+    vetted = if is_nil(user) do nil else user.vetted end
     case get_user_by_id(user_id) do
-      %{super: is_super} -> get_user_data_map(user_id, email, is_super)
-      _ -> get_user_data_map(user_id, email)
+      %{super: is_super} -> get_user_data_map(user_id, email, vetted, is_super)
```
vicmgs, line 22: can we default superuser to vetted

vicmgs (approved):
one comment, lgtm

lib/console/auth/auth.ex
ambuznego: added!

PR #1267, opened 2023-03-17
Blockbax integration

Lynckx:
This PR includes the neccesarry code to integrate helium with the Blockbax platform.  For more information about Blockbax please see our docs: https://blockbax.com/docs/index.html


Lynckx:
In the console when this integration is added the headers are visable in the HTTP details but this contains our API access token which is a secret. How can this be hidden?

PR #1279, opened 2023-08-21
break out events that incur dc charges

michaeldjeffrey:


lib/console/etl_error_worker.ex
```diff
@@ -92,20 +109,20 @@ defmodule Console.EtlErrorWorker do
...
                       }
                     else
                       %{
-                        "dc_balance" => Enum.max([organization.dc_balance - parsed_event.data["dc"]["used"], 0]),
+                        "dc_balance" => Enum.max([organization.dc_balance - event_dc_used(parsed_event), 0]),
                       }
                     end
 
-                  if organization.dc_balance_nonce != parsed_event.data["dc"]["used"] do
+                  if organization.dc_balance_nonce != parsed_event.data["dc"]["nonce"] do
```
michaeldjeffrey, line 125: This was updated, I think it was a typo.

lib/console_web/controllers/router/device_controller.ex
```diff
@@ -455,7 +455,8 @@ defmodule ConsoleWeb.Router.DeviceController do
             cond do
               event["sub_category"] in ["uplink_confirmed", "uplink_unconfirmed"] or event["category"] == "join_request" ->
                 cond do
-                  organization.dc_balance_nonce == event["data"]["dc"]["nonce"] ->
+                  organization.dc_balance_nonce == event["data"]["dc"]["nonce"]
+                    and event["data"]["dc"]["used"] > 0 ->
```
vicmgs, line 459: what was the issue here? why do you need this qualifier?

vicmgs: oh nvm duh i get it

